Issue #83, opened 2025-07-06
CI: tsconfig.test.json not found during build:test step


mickdarling:
## Problem
After fixing the package.json discovery issue in PR #82, we're now hitting a new error:
```
error TS5058: The specified path does not exist: 'tsconfig.test.json'.
```

## Context
- The file exists locally and is tracked in git
- The error occurs when running `npm run build:test` in the diagnostic compile-first approach
- This happens on all platforms (Ubuntu, macOS, Windows)

## Investigation Needed
1. Check if tsconfig.test.json is properly included in the git checkout
2. Verify the working directory is correct when tsc runs
3. Consider if the build:test script needs adjustment for CI environments

## Related
- PR #82 - Fixed package.json discovery
- Issue #81 - CI improvements tracking
- Issue #79 - Original CI failure issue

PR #80, opened 2025-07-06
Fix critical CI path resolution issues (Issues #78, #79)

mickdarling:
## Summary
Fixes critical CI failures that were blocking all PR merges by addressing path resolution issues in Ubuntu and macOS environments.

## Changes
- 🔧 Replaced `__dirname` usage with `process.cwd()` for better CI compatibility
- 🔍 Updated VersionManager to search upward for package.json instead of using fragile relative paths
- 🧹 Removed problematic `fileURLToPath` imports that were causing issues in test environments
- 🐛 Fixed jest configuration by adding `isolatedModules: true` and proper path configurations
- 📊 Added debugging steps to CI workflow to help diagnose file structure issues

## Root Cause
The CI failures were caused by the compiled JavaScript files running from different directory structures in CI than expected locally. The `__dirname` approach was calculating paths relative to the compiled files in `dist/`, but Jest was looking for source files in `src/`.

## Testing
- ✅ All tests pass locally
- 🔄 CI run in progress to verify fixes

## Related Issues
- Fixes #78 (Package.json path resolution in CI environment)
- Fixes #79 (Ubuntu CI failing with ENOENT errors - CRITICAL)

## Next Steps
If CI passes:
1. Re-enable TypeScript build cache in workflow
2. Remove debugging steps from workflow
3. Merge to unblock other PRs

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

PR #86, opened 2025-07-06
Fix CI file deletion by making update classes safer

mickdarling:
## Critical Fix

This PR addresses the root cause of CI failures where files were being deleted after tests ran.

## Problem
After PR #80 changed path resolution to use `process.cwd()`, tests that instantiated BackupManager or UpdateManager could operate on the actual project directory instead of test directories. This caused:
- tsconfig.test.json to be deleted
- package.json to be deleted  
- Other project files to be moved/deleted

## Solution
1. Made `rootDir` configurable in BackupManager and UpdateManager constructors
2. Added safety warning when operating on production directories
3. Tests can now pass safe test directories to avoid operations on real files

## Impact
This should fix:
- Issue #83 (tsconfig.test.json not found)
- The underlying cause of npm ENOENT errors
- Files being deleted during test runs

## Testing
The changes allow tests to provide their own root directory, preventing operations on the actual project files.


mickdarling:
@claude please review this PR again

PR #726, opened 2025-08-25
Fix get_build_info tool return format

mickdarling:
## Summary
Fixes the `get_build_info` tool which was causing Claude Desktop to hang for several minutes due to incorrect return format.

## Problem
The `get_build_info` tool was returning a plain string instead of the required MCP response format. When Claude Desktop called this tool, it would hang waiting for a properly formatted response.

### Root Cause
The handler in `BuildInfoTools.ts` was returning:
```typescript
return buildInfoService.formatBuildInfo(info);  // Returns string
```

But MCP tools require responses in this format:
```typescript
return {
  content: [
    { type: "text", text: "content here" }
  ]
}
```

## Solution
Updated the handler to wrap the formatted build info in the proper MCP response structure.

## Changes Made
- **File**: `src/server/tools/BuildInfoTools.ts`
- **Lines**: 25-36
- Added proper response wrapping with `content` array and `text` type

## Testing
✅ Built the project successfully
✅ Tested the handler directly:
```javascript
// Test output shows correct format:
Response type: object
Has content property: true
Content is array: true
First item type: text
```

## Impact
- ✅ Fixes Claude Desktop hanging when using `get_build_info`
- ✅ Tool now provides useful build information:
  - Package name and version
  - Git commit and branch info
  - Node.js runtime details
  - Memory usage statistics
  - Environment information
  - Server uptime

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

This is a one-line fix that resolves a significant usability issue with the `get_build_info` tool.

### DIFF
--- a/src/server/tools/BuildInfoTools.ts
+++ b/src/server/tools/BuildInfoTools.ts
@@ -24,7 +24,15 @@ export function getBuildInfoTools(_server: IToolHandler) {
       },
       handler: async () => {
         const info = await buildInfoService.getBuildInfo();
-        return buildInfoService.formatBuildInfo(info);
+        const formattedInfo = buildInfoService.formatBuildInfo(info);
+        return {
+          content: [
+            {
+              type: "text",
+              text: formattedInfo
+            }
+          ]
+        };
       }
     }
   ];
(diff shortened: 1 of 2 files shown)


mickdarling:
## ✅ Test Fixes Complete

All failing tests have been updated to expect the correct MCP response format.

### Changes in commit c5225f3:
- Updated 16 test cases in `BuildInfoTools.test.ts`
- Tests now check for MCP format structure: `{ content: [{ type: 'text', text: '...' }] }`
- Extract text content using `result.content[0].text` for assertions
- **All 34 tests now passing** ✅

### CI Status:
The tests should now pass in CI. This completes the fix for the `get_build_info` tool which was causing Claude Desktop to hang.

PR #731, opened 2025-08-25
fix: Add default OAuth Client ID for seamless authentication (v1.6.1)

mickdarling:
## Summary
This PR fixes the GitHub authentication setup experience by including a default OAuth Client ID, enabling "just works" authentication for NPM installs.

## Problem
Users installing DollhouseMCP from NPM had to manually configure an OAuth Client ID before they could use GitHub authentication. This created unnecessary friction:
- `setup_github_auth` would fail with "GitHub OAuth client ID is not configured"
- Users had to create their own OAuth app or find the Client ID
- The intended seamless device flow with 8-character code didn't work out of the box

## Solution
Added DollhouseMCP's official OAuth App Client ID as a default fallback in `GitHubAuthManager`:
- Client ID: `Ov23liXGGP9jNrBhBNfO` (PUBLIC information - only secrets are private)
- Used only when environment variable and config are not set
- Maintains full backward compatibility

## Testing
✅ Tested OAuth flow without any configuration - works perfectly!

## User Experience
**After this fix**:
- Users run: `setup_github_auth`
- Browser opens to: https://github.com/login/device
- Shows 8-character code to enter
- Authentication completes seamlessly

## Release: v1.6.1
This will be released as v1.6.1 - a patch release fixing the authentication setup experience.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docs/development/SESSION_NOTES_2025_08_25_OAUTH_FIX_PR731.md
+++ b/docs/development/SESSION_NOTES_2025_08_25_OAUTH_FIX_PR731.md
@@ -1,0 +1,118 @@
+# Session Notes - August 25, 2025 - OAuth Client ID Fix (PR #731)
+
+**Time**: ~2:00 PM - 3:00 PM  
+**Focus**: Fix GitHub OAuth authentication for NPM installs  
+**Status**: PR #731 created and under review  
+
+## Session Summary
+
+Fixed the "just works" GitHub authentication experience by adding a default OAuth Client ID to the codebase. Users installing from NPM can now use `setup_github_auth` without any configuration.
+
+## Problem Identified
+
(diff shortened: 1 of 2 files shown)


mickdarling:
## ✅ Dead code removed

Thanks for the review! I've removed the unreachable null check as requested.

### Changes in commit 8444a0a:
- **Removed lines 186-193**: The null check that can never be reached
- **Added comment**: Clarifying that `getClientId()` always returns a value

### Testing:
- ✅ TypeScript build successful
- ✅ OAuth flow tested and working
- ✅ Device flow initiation confirmed functional

The code is now cleaner without the unnecessary defensive check. Since `getClientId()` has a guaranteed fallback to the default client ID, this null check was indeed dead code.

Ready for final review! 🚀

PR #1030, opened 2025-09-19
Fix: Memory names showing as 'Unnamed Memory' in list output

mickdarling:
## Summary
This PR fixes the critical bug where all memory elements displayed as "Unnamed Memory" when using the `list_elements` command, even though the memory files had proper names in their YAML frontmatter.

## Problem
When listing memories with `mcp__dollhousemcp-production__list_elements --type memories`, all memories showed as:
```
🧠 Unnamed Memory (v1.0.0) - 
   Retention: permanent days | Tags: none
```

This was happening despite the memory files having valid names like "Conversation History", "Project Context", etc.

### Issue 1: Wrong File Extension
The `MemoryManager.list()` method was looking for `.yaml` files, but memory elements are stored as `.md` files (markdown with YAML frontmatter), similar to personas and other elements.

### Issue 2: Incorrect Property Path  
The `parseMemoryFile()` method was looking for `parsed.metadata` but `SecureYamlParser` returns the YAML frontmatter data in `parsed.data` property.

### 1. File Extension Fix
```typescript
// Before: Only looked for .yaml files
.then(files => files.filter(f => f.endsWith('.yaml')))

// After: Looks for both .md and .yaml files
.then(files => files.filter(f => f.endsWith('.md') || f.endsWith('.yaml')))
```

### 2. Metadata Parsing Fix
```typescript
// Before: Wrong property path
const metadata: MemoryMetadata = {
  name: sanitizeInput(parsed.metadata?.name || 'Unnamed Memory', 100),
  // ...
}

// After: Correct property path
const yamlData = parsed.data || {};
const metadata: MemoryMetadata = {
  name: sanitizeInput(yamlData.name || 'Unnamed Memory', 100),
  // ...
}
```

## Additional Improvements
- Added support for both snake_case and camelCase YAML keys (e.g., `storage_backend` and `storageBackend`)
- Added `parseRetentionDays()` helper to handle various retention formats:
  - Numbers: `30` → 30 days
  - Strings: `"30 days"` → 30 days  
  - Keywords: `"permanent"`, `"perpetual"` → 999999 days
- Improved handling of retention policy structures

## Testing
- ✅ All 89 memory tests pass
- ✅ Build successful with no TypeScript errors
- ✅ Manually verified memory names now display correctly

## Impact
This fix ensures that:
1. Users can see meaningful memory names instead of "Unnamed Memory"
2. Memory metadata is properly parsed from markdown files
3. Both legacy and new memory file formats are supported

## Files Changed
- `src/elements/memories/MemoryManager.ts` - Fixed file extension filtering and metadata parsing
- `docs/development/SESSION_NOTES_*.md` - Added session documentation

## Related Issues
Addresses the memory display issue discovered after v1.9.3 release which fixed memory element support (PR #1028).

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/elements/memories/MemoryManager.ts
+++ b/src/elements/memories/MemoryManager.ts
@@ -645,9 +645,9 @@ export class MemoryManager implements IElementManager<Memory> {
       throw new Error('Invalid file path: Path traversal detected');
     }
     
-    // Ensure proper extension
-    if (!normalized.endsWith('.md') && !normalized.endsWith('.yaml') && !normalized.endsWith('.yml')) {
-      throw new Error('Memory files must have .md, .yaml, or .yml extension');
+    // Ensure proper extension - memories should only be .yaml or .yml
+    if (!normalized.endsWith('.yaml') && !normalized.endsWith('.yml')) {
+      throw new Error('Memory files must have .yaml or .yml extension');
     }
     
     // Construct full path


mickdarling:
## Response to Review Comments

Thank you for the thorough review! You were absolutely correct about the file type confusion.

### ✅ Corrected in Latest Commit

After investigating, I have made the following corrections:

1. **Removed .md file support** - You are right, memories should only be .yaml files
2. **Kept only the real fix** - The actual bug was just the parsed.metadata vs parsed.data issue

### What I Found

The .md files in the root memories directory (conversation-history.md, learning-progress.md, etc.) are:
- Legacy first-pass memory templates/schemas from early development
- Not actual memory storage files
- Should NOT be treated as memories

The real memory files are all .yaml in the date folders, exactly as designed.

### The Actual Fix

The only real bug was the metadata parsing - using parsed.data instead of parsed.metadata since SecureYamlParser returns the YAML frontmatter in the data property.

### About the Legacy .md Files

These files contain valid YAML frontmatter but are memory templates/schemas, not memory instances. They could potentially be:
- Moved to a templates directory
- Converted to .yaml format
- Or simply removed if no longer needed

But that is a separate cleanup task from this bug fix.

### Updated PR Summary

This PR now correctly:
1. Fixes the metadata parsing to use parsed.data
2. Maintains .yaml as the only supported memory file format
3. Includes the helper method for parsing retention policies

Thank you for catching this confusion! The fix is now much cleaner and correct.

mickdarling:
@claude review [7a0048d]

mickdarling:
### Additional Fix Added

Good catch on the validateAndResolvePath() inconsistency! 

I've updated that method as well (line 649-650) to only accept .yaml and .yml extensions, removing the .md support for complete consistency.

Now the entire MemoryManager is consistent that memories are only YAML files.

PR #1032, opened 2025-09-19
Fix: Memory YAML parsing for v1.9.5

mickdarling:
## Summary
This PR fixes the issue where memory names were showing as "Unnamed Memory" even after the v1.9.4 release.

## Problem
Memory files are stored as pure YAML (without frontmatter markers), but the `MemoryManager.load()` method was using `SecureYamlParser` which expects markdown files with YAML frontmatter. This caused the parser to return empty data, resulting in all memories showing as "Unnamed Memory".

## Root Cause Analysis
`SecureYamlParser` is specifically designed for markdown files with YAML frontmatter between `---` markers. When it encounters pure YAML without these markers, it returns `{ data: {}, content: input }`.

The memory files saved by v1.9.3+ are pure YAML with this structure:
```yaml
entries: [...]
metadata: 
  name: "Memory Name"
  description: "..."
extensions: { ... }
stats: { ... }
```

## Solution
1. **Detect format**: Check if content is pure YAML (doesn't start with `---`)
2. **Wrap if needed**: If pure YAML, wrap it with frontmatter markers for SecureYamlParser
3. **Handle nesting**: Update `parseMemoryFile()` to handle nested metadata structure (`data.metadata || data`)
4. **Fix entries loading**: Look for entries in `parsed.data.entries`

## Changes Made
- **MemoryManager.load()**: Added detection for pure YAML vs frontmatter format
- **parseMemoryFile()**: Added fallback logic for metadata location
- **Entries loading**: Fixed to look in the correct location

## Testing
✅ All 36 MemoryManager tests passing
✅ Verified with actual memory files from v1.9.3:
- Memory name correctly parsed as `test-memory-v193` (not "Unnamed Memory")
- Description and other metadata correctly extracted
- Entries loaded successfully

## Impact
This fix ensures that existing memory files created in v1.9.3/v1.9.4 will display their correct names instead of "Unnamed Memory" when listed.

## Related Issues
Fixes the memory name display issue discovered during v1.9.4 testing.

### DIFF
--- a/src/elements/memories/MemoryManager.ts
+++ b/src/elements/memories/MemoryManager.ts
@@ -102,14 +102,20 @@ export class MemoryManager implements IElementManager<Memory> {
       // Memory files are pure YAML (unlike other elements which are markdown with frontmatter)
       // Check if this is pure YAML (doesn't start with frontmatter markers)
       let parsed: any;
-      if (!content.trim().startsWith('---')) {
+      const trimmedContent = content.trim();
+
+      // Handle empty content edge case
+      if (!trimmedContent) {
+        // Empty file - create minimal valid structure
+        parsed = { data: {}, content: '' };
+      } else if (!trimmedContent.startsWith('---')) {
         // Pure YAML file - wrap it with frontmatter markers for SecureYamlParser
         const wrappedContent = `---\n${content}\n---\n`;
         const parseResult = SecureYamlParser.parse(wrappedContent, {
           maxYamlSize: MEMORY_CONSTANTS.MAX_YAML_SIZE,
           validateContent: true
         });
-        // For pure YAML, the entire content becomes the data
+        // For pure YAML, the entire content becomes the data, no markdown content
         parsed = { data: parseResult.data, content: '' };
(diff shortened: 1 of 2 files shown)


mickdarling:
## ✅ Review Feedback Addressed

I've implemented all the suggested improvements from the review:

### Changes Made:

1. **Edge Case Handling** ✅
   - Added check for empty content edge case (line 108-110)
   - Separated empty file logic from pure YAML detection
   - Empty files now gracefully return minimal valid structure

2. **Test Coverage** ✅
   - Added test for pure YAML files without frontmatter
   - Added test for frontmatter format (backward compatibility)
   - Added test for empty file handling
   - Added test for mixed formats in same directory
   - All 40 tests passing (up from 36)

3. **Documentation** ✅
   - Added clarifying comment about data structure
   - Improved code readability

### Test Results:
```
Test Suites: 1 passed, 1 total
Tests:       40 passed, 40 total
```

The fix is now more robust with comprehensive test coverage for all scenarios. Ready for final review!

PR #1716, opened 2026-03-30
fix: log level dropdown defaults to info instead of all

mickdarling:
## Summary
Adds `selected` attribute to the info option in the log viewer dropdown HTML template.

## Problem
PR #1715 set `filterLevel='info'` in JS and synced via `levelSelect.value`, but the browser renders the HTML `<select>` before JS executes. The first `<option>` ("All") was always visually selected on page load.

## Fix
One-line change: `<option value="info" selected>info</option>`

Companion to #1715 (security log backpressure fix). That PR downgraded MEMORY_LOADED events to debug and set the JS filter default. This PR ensures the HTML dropdown matches.

### DIFF
--- a/src/web/public/logs.js
+++ b/src/web/public/logs.js
@@ -100,7 +100,7 @@
             <select id="log-level">
               <option value="">All</option>
               <option value="debug">debug</option>
-              <option value="info">info</option>
+              <option value="info" selected>info</option>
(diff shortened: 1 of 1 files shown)


mickdarling:
Closing — branch contained unrelated commits from auto-dollhouse. Will recreate with a clean cherry-pick.

PR #1801, opened 2026-04-05
feat: token rotation endpoint with TOTP confirmation (#1795)

### LINKED ISSUE #1795: console-auth Phase 2: token rotation endpoint with confirmation

mickdarling:
## Summary

- Adds `POST /api/console/token/rotate` — rotates the primary console token with TOTP second-factor confirmation, returns the new token inline for in-place browser update
- 15-second per-rotation grace window keeps old token valid for in-flight requests from the rotating tab and other local processes
- `DollhouseAuth.refresh(explicitToken?)` extended to accept an explicit token from the rotation response — no page reload needed
- `CONSOLE_TOKEN_ROTATED` security event (HIGH) with no secret material in payload
- Bonus fix: OS verification dialogs now render newlines correctly instead of showing literal `\n` text

## Details

**Store (`consoleToken.ts`):**
- `rotatePrimary(confirmationCode)` — verifies via `verifyTotp()`, stashes old token in in-memory grace buffer, generates new 32-byte token, persists atomically, rebuilds buffer cache
- `verify()` extended to check grace entries with timing-safe comparison after primary tokens
- New `TOTP_REQUIRED` error code for rotation without TOTP enrollment
- Pattern A (OS dialog fallback) deferred — rotation requires TOTP enrollment

**Route (`tokenRoutes.ts`):**
- New router at `/api/console/token` with always-on auth (same pattern as TOTP router)
- Rate-limited: 10 attempts/min
- Structured errors: `MISSING_FIELDS`, `INVALID_TOTP_CODE`, `TOTP_REQUIRED`, `RATE_LIMITED`

**Tests:**
- 12 new store unit tests: happy path, persistence, grace window behavior, concurrent rotations, wrong code, TOTP-not-enrolled, buffer cache rebuild, security event emission + no-leak assertion
- 9 new route integration tests: auth enforcement, missing fields, wrong code, TOTP-not-enrolled, successful rotation, new token authenticates, old token grace, rate limit exhaustion
- 509 web tests total (up from 488 baseline)

**Docs:**
- Token rotation section in `console-auth.md` with endpoint reference, error codes, CLI walkthrough, grace window explanation
- Updated stale Phase 1/Phase 2 references throughout the doc

## Test plan

- [x] `npx jest tests/unit/web/console/consoleToken.test.ts` — 63 tests pass (12 new)
- [x] `npx jest tests/unit/web/routes/tokenRoutes.test.ts` — 9 tests pass (all new)
- [x] `npx jest tests/unit/web/` — 509 tests pass (18 suites, 0 failures)
- [x] `npx tsc --noEmit` — clean typecheck
- [x] Pre-existing test failures (24 suites) confirmed unrelated — same failures on develop baseline

Closes #1795


mickdarling:
@claude please take a look at the latest commit and evaluate it specifically with regards to any noted follow-ups or recommendations any gaps any tests that need to be added or anything else that is lacking in this and evaluate on capability, quality, and approval.

PR #1806, opened 2026-04-06
feat: session auth status labels and always-visible indicator (#1805)

### LINKED ISSUE #1805: Session indicator should show all sessions with auth status labels
## Context

Discovered during #1804 review. The session indicator in the console header only shows sessions connected to the current server instance. When the authenticated console (5907) runs as a standalone `--web` instance, it shows 0 sessions even though the legacy instance (3939) has active MCP sessions. This looks broken from a UX standpoint.

## Problem

1. **Missing sessions**: The authenticated console can't see sessions connected to other DollhouseMCP instances on the same machine
2. **No auth status**: The session dropdown doesn't indicate whether each session is authenticated or unauthenticated
3. **Zero-session state**: When no sessions are visible, the indicator disappears entirely instead of showing the current web session

## What needs to happen

1. **Session auth labeling**: Add an `authenticated` field to the session data model. The dropdown should label each session as authenticated or unauthenticated (e.g., badge or icon).

2. **Cross-instance session visibility**: The authenticated console should aggregate sessions from all DollhouseMCP instances on the machine. Options:
   - Shared session state file under `~/.dollhouse/run/`
   - The authenticated console queries the legacy instance's `/api/sessions` endpoint
   - Leader/follower session forwarding (already partially exists via ingest routes)

3. **Always-visible indicator**: When no MCP sessions are connected, the indicator should still show "1 session" (the current web session) rather than disappearing. Per existing spec.

## Acceptance criteria

- [ ] Session dropdown shows sessions from all local DollhouseMCP instances
- [ ] Each session is labeled as authenticated or unauthenticated
- [ ] At minimum "1 session" always visible (current web session)
- [ ] Works when both legacy (3939) and authenticated (5907) consoles are running

## Refs

- #1804 — token injection fix that exposed this gap
- #1780 — Phase 1 auth infrastructure
- `src/web/public/sessions.js` — session indicator code
- `src/web/server.ts` — `/api/sessions` endpoint

mickdarling:
## Summary

- Session indicator is never empty — a "Web Console" session is always registered
- Each session in the dropdown shows an auth status badge: **auth** (green), **no auth** (red), **console** (indigo)
- New `authenticated` and `kind` fields on `SessionInfo` data model
- Works in both leader mode (UnifiedConsole) and standalone `--web` mode

## Details

**SessionInfo changes (`IngestRoutes.ts`):**
- `authenticated: boolean` — set from `res.locals.tokenEntry` for followers, `true` for leader/console
- `kind: 'mcp' | 'console'` — distinguishes MCP stdio sessions from the web console itself
- `registerConsoleSession()` — new function, creates a persistent console session entry
- Stale reaper skips `kind: 'console'` sessions (they don't send heartbeats)

**Browser (`sessions.js` + `sessions.css`):**
- Auth badge rendered after session name with `data-auth` attribute for styling
- Dark mode support for all three badge states
- Removed `if (count === 0) return` — indicator always renders

**Standalone `--web` mode (`index.ts`):**
- Sets up ingest routes so the session registry exists
- Registers the console session immediately

## Test plan

- [x] `npx tsc --noEmit` — clean typecheck
- [x] `npx jest tests/unit/web/` — 528 tests pass
- [x] Manual: API returns sessions with `authenticated` and `kind` fields
- [x] Manual: "2 sessions" visible in header on 5907

Closes #1805

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -833,8 +833,16 @@ if ((isDirectExecution || isNpxExecution || isCliExecution) && (!isTest || isTes
         metricsSink = new MemoryMetricsSink(240);
       }
 
+      // Set up ingest routes so --web mode has a session registry (#1805).
+      // Without this, the session indicator is always empty in standalone mode.
+      const { createIngestRoutes } = await import('./web/console/IngestRoutes.js');
+      const ingestResult = createIngestRoutes({
+        logBroadcast: (entry) => { /* wired after server starts */ },
+      });
+      ingestResult.registerConsoleSession();
+
       const { startWebServer } = await import('./web/server.js');
-      await startWebServer({ portfolioDir, port, openBrowser: !noBrowser, mcpAqlHandler, memorySink, metricsSink });
+      await startWebServer({ portfolioDir, port, openBrowser: !noBrowser, mcpAqlHandler, memorySink, metricsSink, additionalRouters: [ingestResult.router] });
 
       // Listen for quit commands on stdin (standalone --web mode only).
       // In MCP stdio mode, stdin is consumed by the JSON-RPC transport.
--- a/src/web/console/IngestRoutes.ts
+++ b/src/web/console/IngestRoutes.ts
@@ -51,6 +51,10 @@ export interface SessionInfo {
   lastHeartbeat: string;
   status: 'active' | 'ended';
   isLeader: boolean;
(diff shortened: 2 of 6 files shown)


mickdarling:
@claude we'll Check the latest commits in regard Records to the notes and requests and recommendations, please.

Issue #1850, opened 2026-04-08
bug: Web console tabs empty when launched from Claude Desktop on machine with prior install


mickdarling:
## Description

When the web console is opened via Claude Desktop (MCP stdio mode) using `open_portfolio_browser`, the console opens at `dollhouse.localhost:41715` but all data tabs are empty/non-functional:

- **Auth tab**: Console Token and Authenticator TOTP panels show no content
- **Logs tab**: Empty despite log files existing on disk
- **Metrics tab**: Not working
- **Sessions tab**: Shows no sessions despite being in an active Claude Desktop session
- **Permissions tab**: Reports having a server but no data

## Environment

- Machine with prior DollhouseMCP install (v1.x `.auth/` directory present, v2.x state files from recent use)
- Running `@dollhousemcp/mcp-server@2.0.12-rc.1` via Claude Desktop
- Console opens successfully at `dollhouse.localhost:41715`
- `~/.dollhouse/run/console-token.auth.json` exists (606 bytes) — token store IS initialized
- `~/.dollhouse/logs/` has daily log files through today — file-based logging works
- `~/.dollhouse/portfolio/` has content in all element type folders
- Multiple stale `permission-server-*.port` files in `run/`

## Key Observation

The state files exist on disk and are populated. The HTTP server starts and serves the console UI. But the API routes return empty data. This means the issue is in the **wiring between in-memory sinks and HTTP routes**, not initialization.

## Likely Root Cause

When running through Claude Desktop (MCP stdio mode), the web console is started via `UnifiedConsole.ts` leader/follower architecture. Data flows through in-memory sinks (`MemoryLogSink`, `MemoryMetricsSink`), not disk files. If those sinks aren't connected to the Express routes, the console opens but every tab is empty.

Possible factors:
1. Leader/follower election failing silently — console starts as follower but no leader exists
2. In-memory sinks not being passed to the HTTP route handlers
3. Token store routes mounting but not reading from the initialized `console-token.auth.json`
4. Session registry not receiving registrations from the ingest routes
5. Stale `.auth/` directory from v1.x causing confusion with v2.x auth paths
6. Multiple stale `permission-server-*.port` files interfering with port discovery

## Files to Investigate

- `src/web/console/UnifiedConsole.ts` — leader/follower wiring
- `src/web/server.ts` — route mounting and sink injection
- `src/web/console/consoleToken.ts` — token store data retrieval when `data` may be null
[...]

PR #1851, opened 2026-04-08
fix: empty console tabs when launched from Claude Desktop (#1850)

mickdarling:
## Summary

Fixes three cascading bugs that caused all data tabs (Auth, Logs, Metrics, Sessions) to be empty when the web console was opened from Claude Desktop.

**Root cause chain** (from [diagnostic report](https://github.com/DollhouseMCP/mcp-server/issues/1850)):
1. Leader dies → followers detect via ForwardingSink failures but **don't re-elect**
2. User asks to open console → `openPortfolioBrowser` starts a **bare server** (no sinks) on port 41715
3. Claude Desktop restarts → new leader claims lock, mounts routes in memory, gets **EADDRINUSE** → silently defers to the broken bare server

### Bug A: EADDRINUSE silent deferral (`server.ts`, `UnifiedConsole.ts`)
- `bindAndListen` now returns `BindResult` instead of void
- `handleListenError` logs at WARN (not INFO)
- `startAsLeader` retries port binding up to 3 times with 1s/2s/4s backoff

### Bug B: Bare server startup (`server.ts`, `MCPAQLHandler.ts`)
- Refactored `openPortfolioBrowser` to accept `OpenBrowserOptions` with optional sinks
- Self-provisions `MemoryLogSink`, `MemoryMetricsSink`, `ConsoleTokenStore`, and ingest routes when starting fallback server
- `dispatchBrowser` passes live sinks from `HandlerRegistry`

### Bug C: Follower-to-leader promotion (`LeaderForwardingSink.ts`, `UnifiedConsole.ts`, `LeaderElection.ts`)
- Added `onLeaderDeath` callback to `LeaderForwardingLogSink` (fires via `queueMicrotask` after 5 failures)
- New `promoteToLeader()` function: stops forwarding → deletes stale lock → claims leadership → starts full leader path
- `promotionInProgress` guard prevents concurrent promotions
- Exported `LOCK_VERSION` from `LeaderElection.ts`

## Test plan

- [x] 23 new tests in `console-failure-modes.test.ts` covering:
  - Fresh machine / stale lock / dead PID / expired heartbeat scenarios
  - EADDRINUSE detection via real net.Server conflict
  - ForwardingSink `onLeaderDeath` callback fires after 5 failures
  - Callback fires exactly once even with continued writes
  - Callback does NOT fire when leader is reachable
  - SessionHeartbeat start/stop lifecycle and idempotent double-stop
  - BindResult/WebServerResult type contracts
  - Promotion guard and module exports
- [x] All 710 web tests pass (27 suites, zero regressions)
- [x] Build passes
- [ ] Integration test on fresh machine: start via Claude Desktop, verify all tabs populated
- [ ] Kill leader process, wait 30s, reopen console — should self-heal via promotion

### DIFF
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -197,6 +197,23 @@ const envSchema = z.object({
   DOLLHOUSE_CONSOLE_LEADER_LOCK_FILE: z.string().optional(),
 
   /**
+   * Issue #1850: Retry delays (in ms) when the leader fails to bind the console
+   * port due to EADDRINUSE. Each value is a successive backoff delay.
+   * Default: 1s, 2s, 4s (7s total). Increase for slow or remote environments.
+   */
+  DOLLHOUSE_CONSOLE_BIND_RETRY_DELAYS: z.string()
+    .optional()
+    .transform(v => v ? v.split(',').map(Number).filter(n => !Number.isNaN(n) && n > 0) : undefined),
+
+  /**
+   * Issue #1850: Number of consecutive forwarding failures before a follower
+   * declares the leader dead and attempts self-promotion. Higher values reduce
+   * false positives in high-latency environments but delay recovery.
+   * Default: 10.
+   */
+  DOLLHOUSE_CONSOLE_MAX_FORWARD_FAILURES: z.coerce.number().int().min(1).max(100).default(10),
+
+  /**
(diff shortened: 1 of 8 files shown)


mickdarling:
@claude please look at the latest commits and let me know if there are any outstanding issues, problems, gaps, etc. And evaluate the code that's there for quality, capability, stability.

PR #1855, opened 2026-04-08
fix: kill stale processes squatting on console port (#1850)

mickdarling:
## Summary

The root cause of empty console tabs: stale DollhouseMCP processes from prior sessions hold the port indefinitely. New processes mount all routes in memory but can't bind — the browser connects to the old bare server with no sinks.

Previous fix attempts added retry loops, but retries are useless when the zombie is a healthy Node process that never dies.

## Fix

`bindAndListen` now detects and kills the squatter on EADDRINUSE:

1. `findPidOnPort()` — uses `lsof -ti :PORT` to find who holds the socket
2. Compares against the leader lock file PID — if they differ, it's a squatter
3. `killStaleProcess()` — verifies the process is DollhouseMCP AND owned by the current OS user before sending SIGTERM
4. Waits up to 3s, SIGKILL if needed, then retries the bind

### Safety guards
- Only kills processes whose `ps` command line contains `dollhousemcp` or `mcp-server`
- Only kills processes owned by the current OS user (won't affect other users)
- Never kills a process that matches the lock file's elected leader
- Falls back to "existing instance" if anything fails

### Simplification
Removed the external retry loop from `startAsLeader` and the `retryBind` export — `bindAndListen` handles everything internally now.

## Test plan
- [x] Build passes
- [x] 161 console/port tests pass
- [ ] On Ziggy: kill old zombies, run `npx @dollhousemcp/mcp-server@rc --web`, verify all tabs work
- [ ] Start two sessions, verify leader/follower election still works (follower doesn't get killed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [2.0.11-rc.1] - 2026-04-08
+
+Release candidate for v2.0.11 — console auth, permissions, licensing, port 41715, channel selector
+
 ## [Unreleased] — Phase 2 authenticated console
 
 ### Breaking: Web console default port moved from 3939 → 41715
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dollhousemcp/mcp-server",
-  "version": "2.0.10",
+  "version": "2.0.11-rc.1",
   "lockfileVersion": 3,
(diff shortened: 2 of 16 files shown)


mickdarling:
@claude can you please evaluate this latest commits for this PR, specifically in regards to any comments, recommendations, notes, et cetera?

PR #1860, opened 2026-04-09
fix: pre-flight zombie kill + skip completeDeferredSetup in --web mode (#1850)

mickdarling:
## Summary

Two fixes to the standalone `--web` path that finally resolve the empty console tabs on machines with zombie processes:

1. **Pre-flight zombie kill** — `recoverStalePort()` runs as the first action, before container bootstrap or `startWebServer`. Clears stale processes from the target port immediately.

2. **Skip `completeDeferredSetup()`** — it runs UnifiedConsole leader election which calls `startWebServer()` internally, sets `serverRunning=true`, and causes the `--web` path's own `startWebServer` to early-return without binding. Standalone `--web` IS the server — no leader/follower needed.

## Verified

On Ziggy (machine with zombie rc.1 processes from 12+ hours ago):
- rc.5 killed the squatter (PID 53988)
- Bound to port 41715
- All API routes return live data (logs, health, auth, TOTP)
- Stale port files swept

## Test plan
- [x] Build passes
- [x] 46 unit tests pass
- [x] Verified on real machine with zombie processes

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -848,6 +848,18 @@ if ((isDirectExecution || isNpxExecution || isCliExecution) && (!isTest || isTes
       const cliPort = portArg ? Number.parseInt(portArg.split('=')[1], 10) : undefined;
       const noBrowser = process.argv.includes('--no-open');
 
+      // Pre-flight: kill any stale DollhouseMCP process squatting on our port
+      // BEFORE any container/server setup. This is the definitive fix for #1850 —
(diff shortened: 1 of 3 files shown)


mickdarling:
@claude please review the latest commits specifically in regards to suggestions and issues noted earlier. Note any gaps in testing or capability or stability and anything that It needs to be done for production ready code that's going live immediately.

PR #1903, opened 2026-04-11
fix: auto-apply NVM-aware launcher on install (#1902)

### LINKED ISSUE #1902: fix: NVM PATH ordering causes npx crash on Claude Desktop startup
## Problem

Users with NVM (Node Version Manager) installed can't get the server to start in Claude Desktop, even when pointing to a full path to npx. The error:

```
Error: Cannot find module 'node:path'
Require stack:
- .nvm/versions/node/v24.14.1/lib/node_modules/npm/lib/cli.js
```

**Root cause:** Claude Desktop has its own NVM detection that scans `~/.nvm/versions/node/`, finds all installed versions, and builds a PATH with all of them in ascending order (oldest first). It picks the newest npx binary (e.g. v24) but v12 ends up first in PATH. When npx's `#!/usr/bin/env node` shebang resolves, it finds Node 12 — which doesn't support the `node:` built-in prefix used by npm v24's own CLI. The crash happens before our package is even fetched.

Setting `nvm alias default 24` does not fix it — Claude Desktop never reads NVM aliases; it always scans the versions directory directly. Pointing the config to the full path to npx also doesn't fix it, because the shebang resolution happens against the broken PATH regardless.

The only user-facing workaround today: remove NVM and use Homebrew Node instead. That's too heavy an ask.

## Proposed Fix

During installation (in `runInstallMcp` / `installHandler` in `setupRoutes.ts`), after `install-mcp` writes the config:

1. Detect NVM: check if `~/.nvm/nvm.sh` exists
2. If present:
   - Create `~/.dollhouse/bin/dollhousemcp-launch.sh` (NVM-aware wrapper)
   - Patch the written Claude Desktop config to use the wrapper as `command`, keeping `@dollhousemcp/mcp-server@latest` (or whatever version tag) as the arg

### Wrapper script

```bash
#!/bin/bash
# DollhouseMCP NVM-aware launcher
# Created by the DollhouseMCP installer to work around Claude Desktop's
# broken NVM PATH construction (sorts all installed versions ascending,
# putting oldest first regardless of nvm alias default).

export NVM_DIR="${NVM_DIR:-$HOME/.nvm}"
if [ -s "$NVM_DIR/nvm.sh" ]; then
    . "$NVM_DIR/nvm.sh" 2>/dev/null
    # If the NVM default is < Node 18, upgrade to highest installed version
    MAJOR=$(node -e "process.stdout.write(String(process.versions.node.split('.')[0]))" 2>/dev/null || echo "0")
    if [ "$MAJOR" -lt 18 ]; then
        nvm use node 2>/dev/null || true  # 'node' alias = highest installed
    fi
fi
exec npx "$@"
```

### Config after patching

```json
{
  "command": "/Users/john/.dollhouse/bin/dollhousemcp-launch.sh",
  "args": ["@dollhousemcp/mcp-server@latest"]
}
```

`#!/bin/bash` is always at `/bin/bash` — no PATH ambiguity. NVM gets initialized properly (same as a terminal session). Version tag stays in `args` so users can still edit it manually.

## Implementation Notes

- Detection and patching should happen in `setupRoutes.ts` after the `execFile` call in `runInstallMcp` succeeds
- Wrapper lives in `~/.dollhouse/bin/` — already our owned directory
- Wrapper must be `chmod +x` after creation
- Only created when NVM is detected — no change in behavior for non-NVM users
- If the wrapper already exists (reinstall), overwrite it to pick up any improvements
- Should also handle the re-install case: if the config already points at an old wrapper path, update it

## Affected Users

Any user with NVM installed who runs Claude Desktop. NVM is extremely common among Node developers. This is a silent failure — the server never starts, the error message points at npm internals, and there's no obvious hint that NVM is the cause.

## Confirmed Real-World Report

User reported installing via the one-liner, Claude Desktop failing to start the server, trying `nvm alias default 24`, still failing, trying a full path to npx, still failing. Resolved only by removing NVM entirely.

mickdarling:
## Summary

- Detects NVM presence after a successful `install-mcp` run (checks `~/.nvm/nvm.sh`)
- Creates `~/.dollhouse/bin/dollhousemcp-nvm.sh` — a bash wrapper that sources NVM, upgrades to the highest installed version when the active Node is < 18, then delegates to `exec npx "$@"`
- Patches the written MCP client JSON config to use the wrapper as `command` instead of bare `npx`
- No-ops on Windows and for non-NVM users — zero behaviour change for the common case

## Problem

Claude Desktop has a bug where it scans `~/.nvm/versions/node/`, builds a PATH from all installed versions in **ascending order** (oldest first), picks the newest npx binary, then runs it under that broken PATH. The npx shebang (`#!/usr/bin/env node`) resolves to the oldest installed Node via `env`. This causes a hard crash inside npm's own CLI before our package is ever fetched:

```
Error: Cannot find module 'node:path'   ← npm v24 code, running under Node 12
Require stack:
- .nvm/versions/node/v24.14.1/lib/node_modules/npm/lib/cli.js
```

Setting `nvm alias default 24` doesn't help — Claude Desktop never reads NVM aliases. Pointing the config at the full npx path doesn't help — the shebang still resolves via the broken PATH. The only user-facing fix today is removing NVM entirely.

Closes #1902

## Test plan

- [x] 33 new unit tests in `tests/unit/web/setupRoutes.nvm.test.ts`
  - `isNvmPresent`: no NVM, dir-without-file, file present, empty file
  - `ensureNvmLauncher`: dir creation, return path, executable bits, shebang, NVM sourcing, version check, Node < 18 fallback, `exec npx "$@"` delegation, trailing newline, idempotent overwrite
  - `patchConfigForNvmLauncher`: missing file, malformed JSON, TOML skip, `mcpServers` patch, `servers` patch (VS Code), no dollhousemcp entry, args/keys/siblings preserved, valid JSON output, unknown client, idempotent re-patch
- [x] `tsc --noEmit` clean
- [x] All tests pass


mickdarling:
@claude please evaluate this latest commit specifically in regards to all the notes made in the previous reviews. Thank you.

mickdarling:
@claude please review the latest commits specifically in regards to any notes about enhancements or recommendations. Thank you.

PR #1906, opened 2026-04-11
fix: flush stale LRU cache on ensemble activation (#1895)

### LINKED ISSUE #1895: bug: ensemble activation serves stale cached element list after elements are added
## Summary

When new elements are added to an ensemble definition and the ensemble is deactivated then reactivated, the activation response returns the old element count from a cached version rather than the updated definition.

## Observed Behavior

The DQM-Stack ensemble had 4 new elements added (rank-candidate-keys, match-keys-across-datasets, infer-field-lineage, lineage-discovery-agent), bringing the expected count to 37. After the addition:

1. `activate_element` on the ensemble returned **33 elements** (the pre-update count)
2. `deactivate_element` followed by `activate_element` **still returned 33 elements**
3. The duration on the second activation was **2ms** — confirming it was served from cache, not re-read from disk
4. `get_element_details` on the ensemble also returned the stale 33-element list
5. `search_elements` for the new element names returned 0 results despite the elements existing in the portfolio

**Workaround**: Activate each new element individually via `activate_element` with explicit `element_type` and `element_name`. All 4 elements activated successfully this way and functioned correctly.

## Expected Behavior

- Deactivating and reactivating an ensemble should force a fresh read of the ensemble definition from disk
- `get_element_details` should always reflect the current on-disk state, not a cached version
- Newly added elements should be discoverable via `search_elements` immediately after creation

## Steps to Reproduce

1. Create or load an ensemble with N elements
2. Add new elements to the ensemble definition (edit the YAML)
3. Call `activate_element` on the ensemble
4. Observe: activation reports N elements, not N+4
5. Call `deactivate_element`, then `activate_element` again
6. Observe: still reports N elements, duration ~2ms (cache hit)

## Environment

- Ensemble: DQM-Stack
- Elements added: rank-candidate-keys, match-keys-across-datasets, infer-field-lineage, lineage-discovery-agent
- Expected count: 37 | Actual count returned: 33

## Impact

Medium — ensemble-level cache staleness means newly added elements are silently excluded from ensemble activations. The operator has no indication anything is wrong (0 failures reported, just a wrong count). Requires manual per-element activation as a workaround.

## Suggested Fix Areas

- Ensemble activation should invalidate cache on `deactivate_element`
- Consider a cache TTL or a `force_reload` parameter on `activate_element`
- `get_element_details` for ensembles should bypass cache or have a shorter TTL than activation cache

mickdarling:
## Summary

- `activateEnsemble()` called `findByName()` which hit the in-memory LRU cache first, never triggering a disk scan
- Externally-edited ensemble YAML (e.g. adding new elements) was invisible — activation returned the old element count in ~2ms (cache hit)
- Deactivate + reactivate didn't help; the stale entry stayed in cache

## Root cause

`BaseElementManager.list()` runs `storageLayer.scan()` → detects mtime changes → evicts stale LRU entries. But `findByName()` short-circuits on a cache hit and never calls `list()`. Even when `list()` was called (e.g. in `getElementDetails`), the 1-second scan cooldown would suppress eviction if called in quick succession.

## Fix

Add `BaseElementManager.scanAndEvict()`:
- Calls `storageLayer.invalidate()` to bypass the cooldown
- Runs a fresh mtime-diff scan
- Evicts any modified/removed entries from the LRU

Wire it into `EnsembleManager.activateEnsemble()` so every activation starts with a fresh read of the on-disk definition.

## Test plan
- [x] `should return fresh element count after external file edit` — reproduces the exact scenario from the issue (2-element cache, file updated to 3, activation returns 3)
- [x] `should not return stale data on repeated activate-deactivate-activate cycle` — covers the deactivate+reactivate path described in the issue
- [x] All 43 existing `EnsembleManager` unit tests pass
- [x] Full build compiles cleanly

Closes #1895

### DIFF
--- a/src/elements/base/BaseElementManager.ts
+++ b/src/elements/base/BaseElementManager.ts
@@ -1090,6 +1090,31 @@ export abstract class BaseElementManager<T extends IElement> implements IElement
   }
 
   /**
+   * Force a fresh disk scan and evict any modified/removed entries from the
+   * in-memory LRU cache. Call before findByName() when freshness is critical
+   * (e.g. on ensemble activation) to pick up external file changes that
+   * occurred since the last scan, even if the scan cooldown is still active.
+   *
+   * Unlike list(), this does not load all elements — it only evicts stale ones.
+   * Fixes #1895 (ensemble activation serving stale cached element list).
+   */
+  protected async scanAndEvict(): Promise<void> {
+    this.storageLayer.invalidate(); // bypass cooldown so the next scan hits disk
+    try {
+      const diff = await this.storageLayer.scan();
+      for (const relPath of [...diff.modified, ...diff.removed]) {
+        const absPath = path.join(this.elementDir, relPath);
+        const existingId = this.filePathToId.get(absPath);
+        if (existingId) {
+          this.elements.delete(existingId);
+          this.filePathToId.delete(absPath);
+          this.elementGenerations.delete(existingId);
+        }
+      }
+    } catch { /* non-fatal — cache may be slightly stale, but activation proceeds */ }
+  }
+
+  /**
    * Removes an element from both caches by file path
    * This is the preferred method for deletion to avoid stale cache entries
    * @param filePath - File path (relative or absolute)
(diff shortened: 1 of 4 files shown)


mickdarling:
@claude can you take a look at this in detail and make sure we're not breaking anything? Though we need to add any more tests, are there any other regressions or anything else that could be causing trouble or any gaps? Gaps in how we built this or designed it, thank you.

Issue #99, opened 2025-07-06
Create comprehensive CI troubleshooting guide


mickdarling:
## Context
From PR #89 review - Low Priority recommendation.

## Description
Create a comprehensive troubleshooting guide for CI/CD issues to help contributors quickly diagnose and fix common problems.

### 1. Common CI Failures and Solutions
- **Windows shell syntax errors** - Use `shell: bash` or avoid Unix-specific commands
- **Missing environment variables** - How to set them globally vs per-job
- **Module resolution issues** - ESM vs CommonJS problems and fixes
- **Path issues** - Differences between local and CI environments
- **Timeout failures** - How to increase timeouts or optimize slow tests

### 2. Debugging Techniques
- How to add debug output to workflows
- Using GitHub Actions artifacts to save logs
- Running workflows with debug logging enabled
- Reproducing CI failures locally

### 3. Platform-Specific Issues
- Windows PowerShell vs Bash differences
- macOS-specific quirks
- Linux distribution variations
- Node.js version compatibility

### 4. Quick Reference
- Workflow syntax cheat sheet
- Common environment variables
- Useful GitHub Actions commands
- Links to official documentation

### 5. Best Practices
- Writing cross-platform compatible scripts
- Structuring workflows for maintainability
- Optimizing CI performance
- Security considerations

## Proposed Location
`docs/development/CI_TROUBLESHOOTING_GUIDE.md`

## Format Example
```markdown
# CI Troubleshooting Guide

## Quick Diagnosis Flowchart
Is it failing on Windows only? → Check shell syntax
Is it failing on all platforms? → Check environment variables
...

## Common Issues

### Issue: "command not found: 2>/dev/null"
**Platform**: Windows
**Cause**: PowerShell doesn't support Unix redirects
**Solution**: Add `shell: bash` to the step
...
```

## Benefits
- **Faster problem resolution** for contributors
- **Reduced support burden** on maintainers
- **Knowledge preservation** from past issues
- **Onboarding resource** for new contributors

## Priority
Low - Documentation improvement, not blocking functionality

## Related
- PR #89 (CI fixes and review)
- Issue #88 (Original CI issues)
- All CI-related issues and fixes

Issue #98, opened 2025-07-06
Add performance timing to CI debug output


mickdarling:
## Context
From PR #89 review - Low Priority recommendation.

## Description
Add timing information to CI debug output to help identify performance bottlenecks and slow steps in our workflows.

## Benefits
- **Identify slow steps** in CI pipeline
- **Optimize workflow performance** by focusing on bottlenecks
- **Track performance trends** over time
- **Better resource allocation** understanding

### 1. Simple timing for each debug section:
```bash
START_TIME=$(date +%s)
echo "=== Checking Jest configuration ==="
# ... checks ...
END_TIME=$(date +%s)
echo "Duration: $((END_TIME - START_TIME)) seconds"
```

### 2. Use GitHub Actions built-in timing:
```yaml
- name: Debug file structure
  run: |
    echo "::group::Jest Configuration (timing started)"
    time ls jest.* 2>/dev/null || echo "No jest files found"
    echo "::endgroup::"
```

### 3. Comprehensive timing report:
```bash
# At the end of debug output
echo "=== Performance Summary ==="
echo "Jest check: 0.5s"
echo "Node modules: 2.3s"
echo "File structure: 1.1s"
echo "Total debug time: 3.9s"
```

## Additional Metrics to Consider
- Memory usage at key points
- Disk I/O statistics
- Network timing for package downloads
- Test execution time breakdown

## Integration with Monitoring
- Could export timing data to GitHub Actions annotations
- Consider integration with performance monitoring tools
- Create performance baseline for future comparisons

## Priority
Low - Performance optimization, not critical functionality

## Related
- PR #89 (CI improvements review)
- Issue #94 (Enhanced debug output)
- Issue #88 (CI reliability)

Issue #97, opened 2025-07-06
Add ShellCheck linting to CI workflows


mickdarling:
## Context
From PR #89 review - Low Priority recommendation.

## Description
Add ShellCheck static analysis tool to lint shell scripts and bash commands in our CI workflows. This would help catch shell scripting issues before they cause CI failures.

## Benefits
- **Early detection** of shell syntax errors
- **Cross-platform compatibility** warnings
- **Best practices** enforcement
- **Security** issue detection (command injection, etc.)
- **Educational** - helps team learn shell best practices

## Implementation
1. Add ShellCheck to CI pipeline:
```yaml
- name: Run ShellCheck
  uses: ludeeus/action-shellcheck@master
  with:
    severity: warning
    check-together: yes
    scandir: './scripts'
```

2. Also check inline scripts in workflows:
```yaml
- name: Check workflow scripts
  run: |
    # Extract and check bash scripts from workflow files
    shellcheck .github/workflows/*.yml
```

## Scope
- All `.sh` files in the repository
- Inline bash scripts in GitHub Actions workflows
- Scripts in documentation (if feasible)

## Configuration
Create `.shellcheckrc` for project-specific rules:
```
# Ignore info about shell selection
disable=SC2148

# Allow [ ] instead of [[ ]] for POSIX compatibility
disable=SC2039
```

## Priority
Low - Nice to have for code quality but not blocking

## Related
- PR #89 (CI improvements review)
- Issue #88 (CI reliability)

Issue #96, opened 2025-07-06
Document remaining BackupManager test failure in CI


mickdarling:
## Context
From PR #89 review - Medium Priority recommendation.

## Description
One BackupManager safety test consistently fails in CI environments but passes locally. This needs to be properly documented and tracked.

## Current Failure
```
BackupManager Safety Mechanisms › production directory detection › 
should detect production directory without package.json
```

## Analysis
- The test fails because production directory detection behaves differently in GitHub Actions
- This is a test environment issue, not a code bug
- The test expects certain conditions that aren't met in the CI runner environment

## Tasks
1. **Document the root cause** - Why does this specific test fail in CI?
2. **Determine if the test needs adjustment** - Should we modify the test for CI compatibility?
3. **Create a tracking issue** - If this needs deeper investigation
4. **Consider CI environment detection** - Should tests behave differently in CI?

## Potential Solutions
1. Skip this specific test in CI environments
2. Adjust the test to handle CI directory structures
3. Mock the environment more thoroughly
4. Add CI-specific test variants

## Priority
Medium - This is causing noise in CI results and could mask real failures

## Related
- PR #89 (CI fixes review)
- PR #86 (BackupManager safety mechanisms)
- Issue #88 (CI reliability)

PR #1907, opened 2026-04-11
fix: pin hono 4.12.12 and @hono/node-server 1.19.13 via overrides

mickdarling:
## Summary

- 6 open Dependabot alerts (#94–#99) for `hono` and `@hono/node-server`
- Both are transitive deps from `@modelcontextprotocol/sdk` — not direct dependencies
- Adds npm `overrides` to force the patched versions without bumping the SDK

## Vulnerabilities resolved

| Alert | Package | Issue |
|---|---|---|
| #99 | hono | Non-breaking space prefix bypass in `getCookie()` |
| #98 | hono | Missing cookie name validation in `setCookie()` |
| #97 | hono | Incorrect IP matching in `ipRestriction()` for IPv4-mapped IPv6 |
| #96 | hono | Middleware bypass via repeated slashes in `serveStatic` |
| #95 | hono | Path traversal in `toSSG()` |
| #94 | @hono/node-server | Middleware bypass via repeated slashes in `serveStatic` |

All medium severity. The `serveStatic` bypass (#94/#96) is the most relevant since we serve static files in the web console.

## Approach

Used `overrides` (Option A) rather than upgrading `@modelcontextprotocol/sdk` to keep the change surgical. SDK upgrade can be done separately after review.

## Test plan
- [x] `npm list hono @hono/node-server` confirms both show `overridden` at patched versions
- [x] `npm run build` compiles cleanly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1385,9 +1385,9 @@
       "license": "MIT"
     },
     "node_modules/@hono/node-server": {
-      "version": "1.19.11",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.11.tgz",
-      "integrity": "sha512-dr8/3zEaB+p0D2n/IUrlPF1HZm586qgJNXK1a9fhg/PzdtkK7Ksd5l312tJX2yBuALqDYBlG20QEbayqPyxn+g==",
+      "version": "1.19.13",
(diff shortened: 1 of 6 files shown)


mickdarling:
Closing in favor of #1908 — this branch was accidentally based off fix/1895-ensemble-stale-cache instead of develop, causing unrelated files to appear in the diff.

PR #2009, opened 2026-04-15
Fix Claude Code PreToolUse hook shape

### LINKED ISSUE #2006: Fix Claude Code PreToolUse hook response shape mismatch
## Summary

Claude Code is currently logging repeated hook errors like:

- `PreToolUse:<ToolName> hook error`
- `Hook JSON output validation failed — (root): Invalid input`

The underlying problem is that DollhouseMCP's Claude hook path is still returning the older flat response shape for `claude_code`, while current Claude Code expects `hookSpecificOutput.permissionDecision` for `PreToolUse` hooks.

## What we verified

Current repo state:

- [`scripts/pretooluse-dollhouse.sh`](https://github.com/DollhouseMCP/mcp-server/blob/develop/scripts/pretooluse-dollhouse.sh)
  - posts to `/api/evaluate_permission`
  - echoes the raw JSON response unchanged
- [`src/web/routes/permissionRoutes.ts`](https://github.com/DollhouseMCP/mcp-server/blob/develop/src/web/routes/permissionRoutes.ts)
  - fail-opens with flat JSON like `{ "decision": "allow" }`
- [`src/handlers/mcp-aql/evaluatePermission.ts`](https://github.com/DollhouseMCP/mcp-server/blob/develop/src/handlers/mcp-aql/evaluatePermission.ts)
  - still formats `claude_code` as flat `{ decision: ... }`
  - already formats `codex` using `hookSpecificOutput`

## Why this matters

The hook is effectively fail-open right now, so user work continues, but every tool call can produce noisy validation errors in Claude Code. That creates confusion and makes it look like permission hooks are broken even when tool usage continues.

## Expected Claude Code shape

Current Claude Code `PreToolUse` expects a response shaped like:

```json
{
  "hookSpecificOutput": {
    "hookEventName": "PreToolUse",
    "permissionDecision": "allow" | "deny" | "ask",
    "permissionDecisionReason": "..."
  }
}
```

### Stopgap fix in hook script

Update `scripts/pretooluse-dollhouse.sh` so it:

1. Passes through already-wrapped responses unchanged.
2. Translates flat `{ decision, reason/message }` responses into the Claude Code `hookSpecificOutput` envelope.
3. Fails open silently if the response is malformed.

This is the fastest way to stop the noisy Claude Code validation failures.

### Correct follow-up in server/route layer

Update:

- `src/handlers/mcp-aql/evaluatePermission.ts`
- `src/web/routes/permissionRoutes.ts`

So the Claude Code path is internally consistent too, especially for route-level fail-open responses.

## Acceptance criteria

- Claude Code no longer emits `Hook JSON output validation failed` on normal tool use.
- `scripts/pretooluse-dollhouse.sh` wraps flat responses into Claude Code's expected `hookSpecificOutput` shape.
- Already-wrapped responses pass through unchanged.
- Route-level fail-open responses are also Claude-valid for `platform=claude_code`.
- Regression coverage exists for:
  - flat response translation
  - already-wrapped pass-through
  - malformed response fail-open behavior

## Notes

This should be included in the next permission-hook PR rather than treated as a one-off local config problem.

mickdarling:
## Summary
- normalize Claude Code hook responses in the shared pretooluse bridge so legacy flat responses no longer trigger validation spam
- wrap stale flat Claude and VS Code permission route results into the current hookSpecificOutput shape before returning them
- add regression coverage for route normalization and hook-script translation, pass-through, and malformed-response fail-open behavior

Closes #2006

## Validation
- 
- 
up to date, audited 1026 packages in 744ms

171 packages are looking for funding
  run `npm fund` for details

7 vulnerabilities (1 moderate, 6 high)

To address issues that do not require attention, run:
  npm audit fix

Some issues need review, and may require choosing
a different dependency.

Run `npm audit` for details.
- 
- [jest.config] --experimental-vm-modules not detected; using CJS fallback transform

### DIFF
--- a/scripts/pretooluse-dollhouse.sh
+++ b/scripts/pretooluse-dollhouse.sh
@@ -29,6 +29,33 @@ debug() {
   return 0
 }
 
+normalize_response() {
+  local response="$1"
+
+  case "$HOOK_PLATFORM" in
+    claude_code)
+      echo "$response" | jq -c '
+        if (.hookSpecificOutput.permissionDecision? | type) == "string" then
+          .
+        elif (.decision? | type) == "string" and (.decision | IN("allow", "deny", "ask")) then
+          {
+            hookSpecificOutput: {
+              hookEventName: "PreToolUse",
+              permissionDecision: .decision,
+              permissionDecisionReason: (.reason // .message // "")
+            }
+          }
+        else
+          empty
+        end
+      ' 2>/dev/null
+      ;;
+    *)
+      echo "$response"
+      ;;
+  esac
+}
+
 # Discover the port from the port file
 if [[ -f "$PORT_FILE" ]]; then
   PORT=$(cat "$PORT_FILE" 2>/dev/null)
@@ -92,7 +119,12 @@ while [[ $ATTEMPT -le $MAX_RETRIES ]]; do
 
   if [[ $CURL_EXIT -eq 0 ]] && [[ -n "$RESPONSE" ]]; then
     debug "Response (attempt $((ATTEMPT+1))): $RESPONSE"
-    echo "$RESPONSE"
+    NORMALIZED_RESPONSE=$(normalize_response "$RESPONSE")
+    if [[ -n "$NORMALIZED_RESPONSE" ]]; then
+      echo "$NORMALIZED_RESPONSE"
+    else
+      debug "Malformed response for platform $HOOK_PLATFORM — fail open"
+    fi
     exit 0
   fi
 
--- a/scripts/pretooluse-dollhouse.sh
+++ b/scripts/pretooluse-dollhouse.sh
@@ -54,6 +54,8 @@ normalize_response() {
       echo "$response"
       ;;
   esac
+
+  return 0
 }
 
 # Discover the port from the port file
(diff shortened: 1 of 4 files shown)


mickdarling:
@claude please review the latest commit, specifically in regards to any comments or other issues that may have come earlier and how they have been handled. And any gaps or remaining work that needs to be done

mickdarling:
Worth noting how this work plumbs directly into the broader proposal in #2010 (Universal Hook Runtime).

The `normalize_response()` shim function and the server-side `normalizePermissionResponseForPlatform()` are the seed of the `ResponseShaper` service that #2010 generalizes:

- **This PR**: per-platform branching for `claude_code` + `vscode` inline in route + script
- **#2010**: pulls that out into a `ResponseShaper` service with per-platform renderers (claude_code, codex, cursor, cline, windsurf) plus a graceful-degradation rule for output fields the target platform can't consume at the firing event

The defense-in-depth approach here (normalize at both server and shim) is the right call and should carry forward into the generalized version — the shim's normalization protects against legacy server responses; the server's normalization protects against future client mistakes. Each layer covers the other's failure mode.

Two observations worth considering for this PR (or as follow-ups — they don't block merging):

1. **Aligning fail-open behavior across paths.** The `'allow'` fallback in `normalizePermissionResponseForPlatform` when no decision is extractable is a stronger fail-open than the shim's "empty stdout = no decision." Worth aligning the two paths so behavior is predictable across server-down and server-confused cases.
2. **`CLAUDE_COMPATIBLE_HOOK_PLATFORMS` is the kernel of the full mapping table.** The `Set(['claude_code', 'vscode'])` is exactly the right starting structure. When the #2010 runtime work lands, this generalizes into a per-platform renderer registry covering cursor, codex, cline, windsurf — same shape, just more entries.

Otherwise the PR looks like a clean stopgap that won't be churned by the generalization work — the patterns extracted here are the patterns that scale.

PR #2028, opened 2026-04-16
Make web console leadership version-aware

### LINKED ISSUE #2025: Web console: prefer newest compatible session as leader to avoid older-console regressions
## Summary

The unified web console leader election should prefer the newest compatible DollhouseMCP session instead of simply keeping the first live leader.

Today an older session can remain leader while newer sessions forward logs, controls, and session metadata into that older console. This creates a reverse-compatibility failure mode: the old leader tries to render and control data produced by newer versions, so newer capabilities are missing or misrepresented.

## Problem

Current leader election in `src/web/console/LeaderElection.ts` is based on liveness and heartbeat only.

That means:
- an older session can keep leadership indefinitely if it stays alive
- newer sessions become followers even when they have newer console behavior
- the system depends on older console code understanding newer follower payloads
- new controls, status fields, or behavior can silently disappear in the UI

We are currently seeing exactly that: old leaders are trying to place content, controls, and information coming from newer versions, and the web console is missing things.

## Proposed behavior

Make the web-serving leader the newest **compatible** session.

High-level rules:
- include session/server version in leader-election metadata
- compare compatibility before comparing package version
- if a follower is newer than the current compatible leader, it should be able to take leadership
- equal versions should not flap leadership
- incompatible protocol/schema versions should not automatically take over without an explicit compatibility rule

## Likely implementation areas

- `src/web/console/LeaderElection.ts`
- `src/web/console/UnifiedConsole.ts`
- `src/web/console/IngestRoutes.ts`
- session metadata exposed through `/api/sessions`
- targeted tests for leader takeover, compatibility gating, and non-flapping behavior

## Acceptance criteria

- leader lock/session metadata includes the server version needed for election decisions
- a newer compatible session can replace an older leader
- an older session cannot keep leadership over a newer compatible session just by staying alive
- equal-version sessions do not thrash leadership
- session listings expose enough version information to debug leadership decisions
- tests cover stale-leader, unreachable-leader, and version-preferred takeover paths

## Notes

The key issue is compatibility in the old direction: newer sessions can often tolerate older data, but older leaders cannot reliably render newer controls and state. Leadership should therefore converge toward the newest compatible web console process.

mickdarling:
## Summary

Makes the authenticated web console prefer the newest compatible DollhouseMCP session as leader, and adds cache-busting plus forced reload behavior so open browser tabs move onto the new console code instead of staying on stale assets.

Closes #2025.

## What changed

- add `serverVersion` and `consoleProtocolVersion` to console leader/session metadata
- prefer a newer compatible session over an older live leader during election
- propagate version metadata through follower heartbeats, session registry, and `/api/sessions`
- version-stamp local console CSS/JS/image assets in `index.html`
- serve the HTML shell with `no-cache/no-store` while allowing versioned assets to cache aggressively
- add a client-side forced reload path with cache-busting query params
- trigger forced reload when the browser sees a newer leader version than the one currently loaded

## Why

Older leaders were trying to render controls and state produced by newer sessions. That meant the console could silently miss new UI behavior even when the newest code was already running elsewhere on the machine.

This change makes leadership converge toward the newest compatible console process and ensures already-open browser tabs refresh onto the new asset set instead of holding onto stale JS/CSS.

## Verification

- `npm test -- --runInBand tests/unit/web/console/LeaderElection.test.ts tests/unit/web/console/sessionRegistry.test.ts tests/unit/web/console/console-failure-modes.test.ts`
- `npm test -- --runInBand tests/unit/web/tokenInjection.test.ts tests/unit/web/sessionsReload.test.ts tests/unit/web/cacheBustingSources.test.ts`
- `npx eslint src/web/console/IngestRoutes.ts src/web/console/LeaderElection.ts src/web/console/LeaderForwardingSink.ts src/web/console/PromotionManager.ts src/web/server.ts src/web/public/app.js src/web/public/sessions.js tests/integration/console-lifecycle.test.ts tests/unit/web/console/LeaderElection.test.ts tests/unit/web/console/console-failure-modes.test.ts tests/unit/web/console/sessionRegistry.test.ts tests/unit/web/tokenInjection.test.ts tests/unit/web/sessionsReload.test.ts tests/unit/web/cacheBustingSources.test.ts`

### DIFF
--- a/docs/architecture/ADR-002-CONSOLE-VERSION-AWARE-LEADERSHIP.md
+++ b/docs/architecture/ADR-002-CONSOLE-VERSION-AWARE-LEADERSHIP.md
@@ -1,0 +1,72 @@
+# ADR-002: Version-Aware Web Console Leadership
+
+**Status:** Accepted
+**Date:** 2026-04-16
+**Authors:** DollhouseMCP Team
+
+## Context
+
+The authenticated web console can be open in multiple browser tabs while multiple `mcp-server` processes are also running on the same machine. Before this change, leadership was primarily based on liveness and lock ownership. That allowed an older process to remain leader even after a newer process started.
+
+That mismatch created two concrete problems:
+
+1. An older leader could keep serving stale UI assets and session metadata after a newer server was available.
+2. Browser tabs connected to that older leader could miss controls, rendering paths, or session fields introduced by the newer server.
+
+The console needed a deterministic rule for choosing which process should lead, plus a browser-side recovery path so already-open tabs would reconnect to the newest compatible leader.
+
+## Decision
+
+We make web console leadership **version-aware and compatibility-aware**.
+
+### Leader selection
+
+- Each leader lock and session heartbeat now includes:
+  - `serverVersion`
+  - `consoleProtocolVersion`
+- A candidate may replace the current leader only when:
+  - both processes are compatible at the console protocol level, and
+  - the candidate's package version is strictly newer
+- Equal versions do not trigger takeover, which avoids leadership flapping.
+- Leaders without version/protocol metadata are treated as legacy and therefore older than any authenticated-console leader that includes the new fields.
+
+### Browser recovery
+
+- The HTML shell is served with `no-cache, no-store, must-revalidate`.
+- Local CSS, JS, and image assets are stamped with `?v=<asset version>`.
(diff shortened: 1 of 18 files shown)


mickdarling:
@claude please review the latest commit specifically in regards to any comments or reviews or notes previously.

PR #2032, opened 2026-04-16
Hotfix console leader bind fallback and session registration

mickdarling:
## Summary
- make console leadership provisional until the HTTP listener actually binds
- fall back to follower mode and register with the real port owner when bind fails
- harden stale-port recovery so we do not evict live Claude/Codex-hosted MCP servers, and bump release metadata to `2.0.20`

## Testing
- `npx eslint src/web/console/UnifiedConsole.ts src/web/console/StaleProcessRecovery.ts tests/unit/web/console/UnifiedConsole.test.ts tests/unit/web/console/stale-process-recovery.test.ts`
- `npm test -- --runInBand tests/unit/web/console/UnifiedConsole.test.ts tests/unit/web/console/stale-process-recovery.test.ts tests/unit/web/console/console-failure-modes.test.ts tests/unit/web/console/LeaderElection.test.ts tests/unit/web/console/sessionRegistry.test.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [2.0.20] - 2026-04-16
+
+Point release for console leader bind authority and follower registration recovery.
+
 ## [2.0.19] - 2026-04-16
 
 Point release for deadlock relief recovery and version-aware web console leadership.
--- a/manifest.json
+++ b/manifest.json
@@ -2,7 +2,7 @@
   "manifest_version": "0.3",
   "name": "dollhousemcp",
   "display_name": "DollhouseMCP",
-  "version": "2.0.19",
+  "version": "2.0.20",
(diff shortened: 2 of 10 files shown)


mickdarling:
@claude Please do an updated review of the latest commits on this PR head (`590c8138`) and focus specifically on the remaining judgment-call items from the prior review:

- whether `discoverLeaderServingPort()` should add an explicit fetch timeout / AbortController for leader discovery
- whether the current process command-line / parent-command matching in `StaleProcessRecovery.ts` still has any meaningful bypass risk worth fixing in this hotfix
- whether the current lock recheck / TOCTOU mitigation is sufficient for this release
- whether any PID validation or PID-recycling concern remains after the current `parsePidToken()` checks
- whether there are any real edge-case bugs still worth addressing before merge

Context: security audit is now passing and SonarCloud quality gate is passing, so please review against the latest code rather than the earlier findings.

PR #2035, opened 2026-04-16
Replace legacy console leaders on bind recovery

mickdarling:
## Summary
- allow a newer authenticated console to evict an older incompatible Dollhouse leader that still owns the console port
- keep the active-host-parent safety guard for ordinary stale recovery, but bypass it intentionally only after bind recovery proves the discovered leader is older or legacy
- add regression coverage for the split-brain case where a live legacy leader keeps serving the old console and blocks new session registration

## Why
A live `2.0.12-rc.15` leader on port `41715` can currently survive behind the host-parent protection and force a `2.0.21` session to fall back to follower mode. That leaves the browser on the old console and the new session never appears in `/api/sessions`.

## Verification
- `npx eslint src/web/console/UnifiedConsole.ts src/web/console/StaleProcessRecovery.ts tests/unit/web/console/UnifiedConsole.test.ts`
- `npm test -- --runInBand tests/unit/web/console/UnifiedConsole.test.ts tests/unit/web/console/LeaderElection.test.ts tests/unit/web/console/stale-process-recovery.test.ts`

### DIFF
--- a/src/web/console/StaleProcessRecovery.ts
+++ b/src/web/console/StaleProcessRecovery.ts
@@ -87,6 +87,10 @@ export interface KillStaleProcessOutcome {
   detail?: string;
 }
 
+export interface KillStaleProcessOptions {
+  allowActiveHostParent?: boolean;
+}
+
 export function isRecognizedMcpHostParent(command: string): boolean {
   const normalizedCommand = UnicodeValidator.normalize(command).normalizedContent;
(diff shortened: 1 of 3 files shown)


mickdarling:
@claude please do a review on the latest commits specifically with regards to any recommendations or suggestions or potential issues from earlier reviews. Thank you.

PR #2037, opened 2026-04-16
hotfix: converge console authority to the active listener

mickdarling:
## Summary
- fix listener detection so forced takeover targets the actual process bound to the console port instead of any connected client PID
- detect split-brain authority when the lock holder and the active port owner disagree, then prefer the actual port owner or force a newer session to replace an older serving leader
- add focused unit and integration coverage for both the connected-client PID trap and the split-brain follower resolution path

## Why
On the live machine, a newer session could win or rewrite the auth lock while an older process still served `127.0.0.1:41715`. Because the follower path only checked whether the port was reachable, it stayed follower even when the reachable console belonged to the wrong process and served stale assets (`2.0.17`).

## Verification
- `npm test -- --runInBand tests/unit/web/console/UnifiedConsole.test.ts tests/unit/web/console/stale-process-recovery.test.ts`
- `npm run test:integration -- --runInBand tests/integration/console-lifecycle.test.ts -t 'does not mistake connected clients for the listening owner|finds a PID using available system commands'`
- `npx eslint src/web/console/UnifiedConsole.ts src/web/console/StaleProcessRecovery.ts tests/unit/web/console/UnifiedConsole.test.ts tests/integration/console-lifecycle.test.ts`

## Local reproduction notes
- reproduced on the current machine with lock holder `64270` (`2.0.23`) while listener `85513` still served `2.0.17` asset URLs
- verified the old listener was selected correctly once `findPidOnPort()` was constrained to `LISTEN` sockets only

### DIFF
--- a/src/web/console/StaleProcessRecovery.ts
+++ b/src/web/console/StaleProcessRecovery.ts
@@ -298,19 +298,34 @@ export async function findPidOnPort(port: number): Promise<number | null> {
   const { promisify } = await import('node:util');
   const execFileAsync = promisify(execFileCb);
 
-  // Try lsof first (macOS + most Linux), fall back to fuser (minimal Linux/Docker)
+  // Query only LISTEN sockets so established client connections to the console
+  // don't get mistaken for the owning leader process.
   for (const cmd of [
-    { bin: 'lsof', args: ['-ti', `:${port}`] },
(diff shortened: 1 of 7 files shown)


mickdarling:
@claude please review the latest commits specifically in regards to any of the comments or reviews or suggestions that were made in the previous commits.

PR #2056, opened 2026-04-17
Polish authority mode permissions UX

### LINKED ISSUE #2055: Polish authority-mode UX and investigate session-scoped authority
## Summary
Follow-up UX/design pass for the new Permission Authority controls in the web console, plus a design investigation into whether authority can be expressed per session instead of only per platform.

This issue is intentionally a new follow-up after `#2054`, not a retroactive expansion of that PR.

## Context
The first authority-mode UI landed in the Permissions tab in `#2054`, and live preview review surfaced a few important usability problems:

1. The disabled `Authoritative` option for unsupported hosts is too subtle.
2. The explanatory copy for unsupported hosts is too far away from the disabled control.
3. The need to explicitly save a changed selection is too easy to miss.
4. The confirmation/save flow feels a little too quiet relative to the impact of the action.
5. There is a product question about whether authority should be scoped by **session** rather than only by **platform**.

### 1. Make disabled `Authoritative` state more obvious
When `Authoritative` is disabled for a host like Codex:
- increase visual distinction of the disabled state
- make it obvious that the option is intentionally unavailable, not just inert
- consider label styling, lock/badge iconography, or stronger disabled affordance

### 2. Move the unsupported-host explanation next to the disabled control
Current copy:
> Authoritative settings sync currently ships with Claude Code only. Other hosts can use Off or Shared mode for now.

This should live directly adjacent to the disabled `Authoritative` option, not down near the reason/save controls, so the user understands *why that specific control is unavailable*.

### 3. Make unsaved changes much more obvious
Right now it is too easy to click a radio option and assume the change is complete.

Design a clearer unsaved-changes treatment so the user immediately understands:
- the selection changed
- it is not active yet
- a save/confirm action is still required

Examples to consider:
- prominent "Unsaved changes" state near the control group
- stronger save button styling once dirty
- sticky footer or action bar within the card
- inline warning/notice tied to the changed state
- visual differentiation between "current mode" and "selected mode"

Use a strong web-console design pass here rather than a purely utilitarian patch. The save affordance should feel intentional and obvious.

### 4. Review confirmation affordance / tone
The current confirm/save flow is acceptable but still needs polish:
- destructive/scary actions should feel appropriately weighty
- routine mode flips should still feel clear and deliberate
- copy and visual treatment should help the user trust what is about to happen

### 5. Investigate session-scoped authority instead of platform-only authority
Question raised during live review:
- can authority be scoped to a specific live session (`Bernie`, `Kukla`, `Ollie`, etc.) instead of only to a platform (`codex`, `claude-code`, etc.)?

This issue should investigate and document:
- whether `off/shared/authoritative` can realistically be set per session
- how that interacts with host-native settings files, which are usually platform/user scoped
- whether session-scoped authority is feasible for:
  - `off`
  - `shared`
  - `authoritative`
- whether the answer differs between hook behavior and host-settings sync

### Likely constraint to validate
A likely limitation is that `authoritative` sync writes into host-native config that is global or per-project for the host, not per named Dollhouse session. If true, then:
- session-scoped authority may be feasible for hook-side behavior
- but not truly feasible for host-settings-authoritative sync without deeper host support

That needs to be validated and documented clearly.

## Desired outcome
- improved human-usable authority controls in the Permissions tab
- clearer disabled-state messaging
- stronger unsaved-changes affordance
- documented answer on session-vs-platform scoping
- if session-scoped authority is not viable, explain why in code/docs/issue notes so the limitation is explicit

## Acceptance criteria
- [ ] Disabled `Authoritative` state is visually obvious for unsupported hosts
- [ ] Unsupported-host explanatory copy sits adjacent to the disabled `Authoritative` option
- [ ] Dirty/unsaved state is obvious enough that a user is unlikely to mistake selection for completion
- [ ] Save action has a stronger, more intentional visual treatment when a change is pending
- [ ] Confirmation flow/copy gets a UX polish pass
- [ ] Session-vs-platform authority feasibility is investigated and documented
- [ ] If session-scoped authority is not viable, the reason is captured explicitly in the issue/PR/docs

## Related
- `#2053` — permission-authority modes for host vs Dollhouse control
- `#2054` — initial implementation of authority modes and Permissions-tab controls

mickdarling:
## Summary
- make the authority mode options read like explicit selectable cards instead of plain radio rows
- move the unsupported Authoritative host note next to the Authoritative option itself
- add a high-visibility unsaved draft state with host-specific save messaging and stronger save affordances
- extend the UI regression tests to cover unsupported hosts and dirty-save behavior

## Testing
- npm test -- --runInBand tests/unit/web/console-ui-regressions.test.ts
- npx eslint src/web/public/permissions.js tests/unit/web/console-ui-regressions.test.ts

Closes #2055

### DIFF
--- a/src/utils/permissionAuthority.ts
+++ b/src/utils/permissionAuthority.ts
@@ -3,9 +3,14 @@ import { dirname, join } from 'node:path';
 import { mkdir, readFile, rename, rm, writeFile } from 'node:fs/promises';
 import { existsSync } from 'node:fs';
 import { SecurityMonitor } from '../security/securityMonitor.js';
+import { UnicodeValidator } from '../security/validators/unicodeValidator.js';
 import { logger } from './logger.js';
 import { getClaudeHookSettingsPath } from './permissionHooks.js';
 
+// These modes intentionally describe who has the final say when Dollhouse and the host disagree:
+// - off: the host permission system is fully in charge
+// - shared: both systems participate, but the host can still be stricter
+// - authoritative: Dollhouse is the source of truth for the managed host slice
 export const PERMISSION_AUTHORITY_MODES = ['off', 'shared', 'authoritative'] as const;
 export type PermissionAuthorityMode = typeof PERMISSION_AUTHORITY_MODES[number];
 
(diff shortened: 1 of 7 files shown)


mickdarling:
@claude can you please evaluate this latest commit for any issues that were requested in previous reviews?

PR #2059, opened 2026-04-17
Add first-class session identity surfaces

### LINKED ISSUE #2049: [enhancement] First-class session identity: deterministic DOLLHOUSE_SESSION_ID + queryable + displayable
## Summary
`DOLLHOUSE_SESSION_ID` is currently a second-class concept: it exists internally (`ActivationStore.resolveSessionId()` at [src/services/ActivationStore.ts:89-107](../blob/develop/src/services/ActivationStore.ts)), but it's invisible from outside the process, not queryable via any MCP operation, not surfaced in UI, and defaults to a throwaway random ID when unset. This issue promotes it to a first-class identity with three facets:

1. **Deterministic** — unset defaults should produce a stable, restart-surviving ID (not a random throwaway).
2. **Queryable** — any session can be asked "what's your Dollhouse name?" via MCP introspection.
3. **Displayable** — the ID surfaces in status-bar, menu-bar, and similar UI so humans can distinguish a dozen concurrent sessions at a glance.

All three facets are facets of the same concept — solving them together is cleaner than separate issues.

## Motivation
- **Operations**: running a dozen concurrent sessions (mix of Claude Code, Codex, bridges, etc.) is real. Without a visible identity, operators can't tell which is which.
- **Persistence**: `ActivationStore` and future `start_agent_from_handoff` ([auto-dollhouse#96](https://github.com/DollhouseMCP/auto-dollhouse/issues/96)) depend on the ID being deterministic and restart-stable.
- **Event dispatcher plugin** ([auto-dollhouse#97](https://github.com/DollhouseMCP/auto-dollhouse/issues/97)): per-channel session IDs are the deconfliction mechanism ([auto-dollhouse#94](https://github.com/DollhouseMCP/auto-dollhouse/issues/94)). Assumes the ID is a real first-class concept.
- **Hook runtime** ([#2010](https://github.com/DollhouseMCP/mcp-server/issues/2010)): hook-activation persistence across host restarts requires the ID to be deterministic.

## Concrete motivating example — Claude Code statusline

The Claude Code statusline (`~/.claude/statusline-enhanced.sh`) already tries to display the Dollhouse session name, but the mechanism is fragile and **keeps breaking**. It has to:

1. Read the Claude Code parent process's PID from hook input (`jq -r '.session_id'`).
2. Walk the process descendants: `claude (PPID) → npm exec @dollhousemcp/mcp-server → node .../dollhousemcp`.
3. Collect **all candidate PIDs** at every level (the `dollhousemcp` process can be a direct child OR a grandchild spawned through an `npm exec` wrapper — the script comments explicitly call this out).
4. `curl -s http://localhost:41715/api/sessions`
5. JSON-filter the sessions list to find one whose `pid` matches any of the collected candidates.
6. Extract `displayName`, `color`, `isLeader`, cache the result keyed by Claude Code session_id.

Every step above is a failure point:

- Process-tree shape changes (new `npm exec` wrapper behavior, different spawn parents, multi-process launchers) break the PID heuristic.
- The HTTP port `41715` is discovered via the permission-server port file but baked into some paths as a constant.
- `/api/sessions` is an informal non-MCP endpoint; its shape can shift without notice.
- Cache is locked for the Claude Code session_id's lifetime — if the mapping changes mid-session, you can't refresh.

**All of this collapses to one MCP tool call once Facet 2 lands:**

```bash
# Today (fragile)
pids=$(walk_process_tree "$claude_pid")
session=$(curl -s "http://localhost:$port/api/sessions" | jq "...filter by pid...")

# After Facet 2
session_id=$(mcp_call get_build_info | jq -r .sessionId)
```

This is the canonical real-world motivation for Facets 2 and 3 — a live integration that's visibly broken today because there's no clean way to ask a session its name.

### Current behavior
`ActivationStore.resolveSessionId()` generates a per-process random ID (`session-<timestamp>-<hex>`) when `DOLLHOUSE_SESSION_ID` is unset. Most MCP hosts (Claude Code, Claude Desktop, Cursor, etc.) don't set this env var, so:

- Every server launch gets a fresh random ID.
- Activations persist to a file that will never be restored from.
- State directory accumulates orphan `activations-session-<timestamp>-<hex>.json` files with no cleanup pathway.

### Reproduction (@dollhousemcp/mcp-server v2.0.23)
1. Activate an ensemble (e.g., `dqm-stack-v2` — 40 elements).
2. Check `~/.dollhouse/state/activations-*.json` — find a recently-modified file matching the random session ID pattern.
3. Check `~/.dollhouse/state/activations-default.json` — **empty** (all element-type arrays are `[]`).
4. Restart the MCP server. ActivationStore generates a *new* random session ID. The prior file is orphaned.

Observed in my state dir: ~30 orphan `activations-session-*.json` files accumulated.

### Possible fixes
- **A. Auto-derive from transport identity.** Unset → use a stable ID like `mcp-<client-name>-<portfolio-path-hash>` so Claude Code sessions for the same user keep reusing the same file. Requires plumbing client-identification info into ActivationStore construction.
- **B. Default to `'default'` when unset.** Trade-off: cross-session leakage if two concurrent servers run under the same user. For single-server default case, restore-on-startup just works.
- **C. Opt-in persistence explicitly.** Treat `DOLLHOUSE_ACTIVATION_PERSISTENCE=true` as the switch instead of letting random IDs silently write files that will never be read. Write nothing when `DOLLHOUSE_SESSION_ID` is unset.
- **D. Periodic cleanup of orphan files.** Sweep state dir at startup, remove files older than N days whose session IDs won't match. Janitorial only.
- **E. Document explicit-set as the supported configuration.** Pair with `npx @dollhousemcp/mcp-server@latest --web` setup wizard detecting and adding `DOLLHOUSE_SESSION_ID` to the MCP host config.

## Facet 2 — Queryable via MCP

Any session should be askable "what's your Dollhouse name?" — both for programmatic integration and for ad-hoc operator queries from a session drop-down.

### What exists today (informal)
`GET http://localhost:<port>/api/sessions` — an informal HTTP endpoint exposed by the web console. Returns a list of registered sessions with `displayName`, `color`, `isLeader`, `pid`, and `kind`. The Claude Code statusline already depends on this (see the motivating example above). But it's:

- Not an MCP operation (can't be called from any MCP client without port discovery and HTTP plumbing).
- Not versioned / documented as a stable external surface.
- Requires mapping a caller's process PID back to a session entry — the statusline does this via process-tree walking, which is inherently fragile.
- Only accessible same-host; no cross-process identity delivery.

Facet 2 should either **formally bless** `/api/sessions` as a documented external surface AND provide the equivalent data via MCP, or **deprecate** `/api/sessions` in favor of MCP-only access. Either way, removing the PID-walk requirement is the win.

### Options
- **Add `sessionId` to `get_build_info`** — simplest. Existing operation, already surfaces runtime info. One new field: `"sessionId": "zulip-bridge"`. Minimal code change, no new tool registration. Answers the common case ("what session am I talking to right now?") in one MCP call.
- **Dedicated `get_session_identity` operation** — richer: also return `sessionStartedAt`, `portfolioPath`, `mcpClientName` (if known), `activeElementCount`, `sessionRole: leader|follower`. Registered on the READ endpoint. More intentional API surface.
- **Keep `/api/sessions` as the "list all sessions" surface** (distinct from self-identification) — but document and version it.
- **All three** — `get_build_info` for casual self-id, `get_session_identity` for richer self-id records, `/api/sessions` for cross-session enumeration by the web console and the statusline.

### Recommendation
All three. `get_build_info` already runs at session orientation time (it's how integrators check they're connected to the right server); adding `sessionId` costs ~1 line. A dedicated `get_session_identity` gives a clean namespaced path for richer identity data. `/api/sessions` stays for the legitimate cross-session use case (a web console listing all MCP sessions the user has running) but becomes documented/versioned. The statusline migrates to `get_build_info` for the *self*-id case, eliminating the PID-walk.

### Where to surface
- **DollhouseMCP web console** (menu-bar / browser tab): current session indicator with the ID visible at all times.
- **MCP host session pickers** (the drop-down of concurrent sessions Mick sees): the ID should appear in the title/description. This is host-side — requires the ID to be queryable (Facet 2) so host UIs can fetch it.
- **Session picker in auto-dollhouse drawing room**: `LeaderPageEventProxy` already differentiates leader/follower; add session ID to the display.
- **Status bar (Claude Code)**: statusline displays session identity alongside model, uptime, etc.

### Implementation dependencies
Facet 2 (queryable) is a hard prerequisite for Facet 3 (displayable) to work at the host level. The web console (same-process) can read the ID directly.

## Facet 4 — Cross-session disambiguation in ops

When multiple sessions collide on an agent name (auto-dollhouse#94), the 409 Conflict response should include both session IDs (`owner_session_id` and the requester's own ID) so operators can see the collision clearly. This is an extension of Facet 2 — the identity is useful not just for self-identification but for describing other sessions.

## Weak recommendation for implementation order
1. **Fix the deterministic default** (Facet 1, fix A or B) — this is the blocker for everything else.
2. **Add `sessionId` to `get_build_info`** (Facet 2, minimal slice) — unblocks display work.
3. **Surface in the web console** (Facet 3, same-process) — immediate UX win.
4. **Dedicated `get_session_identity` operation** (Facet 2, richer slice) — when host integrations ask for it.
5. **Host-side UI integration** (Facet 3, host-specific) — via host extensions that consume the MCP introspection.

## Related
- [#2010](https://github.com/DollhouseMCP/mcp-server/issues/2010) — universal hook runtime. Hook activation persistence depends on Facet 1 being fixed. Strongest coupling.
- [auto-dollhouse#97](https://github.com/DollhouseMCP/auto-dollhouse/issues/97) — event-dispatcher plugin epic. §5 (deconfliction) relies on first-class session ID.
- [auto-dollhouse#94](https://github.com/DollhouseMCP/auto-dollhouse/issues/94) — agent-name deconfliction. 409 Conflict response needs to include session IDs (Facet 4).
- [auto-dollhouse#96](https://github.com/DollhouseMCP/auto-dollhouse/issues/96) — `start_agent_from_handoff`. Spawn-with-handoff flow takes a `sessionHint`; requires the concept to be deterministic.
- [bridge#149](https://github.com/DollhouseMCP/bridge/issues/149) — `DOLLHOUSE_SESSION_ID` not propagated in bridge's `_run_claude()`. Different repo, related symptom.

mickdarling:
## Summary
- add a deterministic session identity helper that derives a stable `DOLLHOUSE_SESSION_ID` fallback from workspace context
- split stable persistence identity from runtime-unique live console identity to avoid session registry collisions
- expose session identity through `get_build_info` structured content and the web console header

## Details
- `ActivationStore` now keeps a stable session id for persisted activation state and a runtime session id for live console registration
- `get_build_info` and schema-dispatch build info responses now return machine-readable session identity alongside the existing markdown
- the web server injects session meta tags and the console header shows the stable session id for quick operator visibility

## Verification
- `npm test -- --runInBand tests/unit/services/sessionIdentity.test.ts tests/unit/services/ActivationStore.test.ts tests/unit/services/BuildInfoService.test.ts tests/unit/server/tools/BuildInfoTools.test.ts`
- `npx eslint src/di/Container.ts src/index.ts src/services/sessionIdentity.ts src/services/ActivationStore.ts src/services/BuildInfoService.ts src/server/tools/BuildInfoTools.ts src/handlers/mcp-aql/SchemaDispatcher.ts src/web/server.ts src/web/console/UnifiedConsole.ts src/web/public/app.js tests/unit/services/sessionIdentity.test.ts tests/unit/services/ActivationStore.test.ts tests/unit/services/BuildInfoService.test.ts tests/unit/server/tools/BuildInfoTools.test.ts`
- `npm run build`

Closes #2049.


mickdarling:
@claude please review the latest commits specifically with regards to any notes from the earlier reviews. Note code quality, readiness for release, any gaps. or anything else worth mentioning.

PR #2062, opened 2026-04-17
Fix session meta injection for direct index shell requests

mickdarling:
## Summary
- route direct /index.html requests through the injected SPA shell instead of raw express.static output
- keep static assets on express.static while skipping HTML shell files
- add regression coverage for session/runtime session/asset placeholder injection on both / and /index.html

## Testing
- npm test -- --runInBand tests/unit/web/tokenInjection.test.ts
- npx eslint src/web/server.ts tests/unit/web/tokenInjection.test.ts


mickdarling:
@claude please review this pr

PR #2102, opened 2026-04-19
test: add mixed-version console harness

mickdarling:
## Summary
- add a dedicated Docker harness for mixed-version console leader-election debugging
- include live observer and markdown report generation for lock/session state
- add an injection helper so one running service can be swapped to a different package version without editing compose files

## What I validated
- brought up the full harness with current-local plus historical published versions
- captured live mixed-version state with `observe-state.sh`
- generated timestamped reports with `report-state.sh`
- injected a version change with `inject-service.sh legacy-225 @dollhousemcp/mcp-server@2.0.24`
- confirmed the shared session view updated to show the injected follower on `2.0.24`

## Notes
- the observer is the source of truth because it queries `/api/sessions` from inside the shared Docker namespace
- the host-facing published port is best-effort only because the console binds loopback inside that namespace

### DIFF
--- a/.github/workflows/docker-testing.yml
+++ b/.github/workflows/docker-testing.yml
@@ -647,6 +647,17 @@ jobs:
         with:
           fetch-depth: 1
 
+      - name: Install shellcheck
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y shellcheck
+
+      - name: Validate mixed-version harness shell scripts
+        shell: bash
+        run: |
+          shellcheck docker/mixed-version/*.sh
+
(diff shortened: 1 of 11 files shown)


mickdarling:
@claude please re-review the latest commits on this PR, specifically the harness hardening work in the most recent updates: the non-root Dockerfile/runtime changes, the explicit Docker build input allowlist, the observer fallback rewrite, the package-spec validation, and the shared-workspace install lock. The previous review appears to be based on an older snapshot, so please focus on the current HEAD and call out only any remaining actionable issues in that latest state.

PR #2112, opened 2026-04-20
Implement leader-authoritative session display-name leasing

### LINKED ISSUE #2111: Define ideal session display-name leasing architecture for unified console
## Summary
Define the ideal session display-name architecture for the unified web console so mixed-version, reconnecting, and leader-takeover scenarios do not rename live sessions, duplicate names, or let leaders invent conflicting names.

## Problem
We have been fixing several symptoms in the console leadership/session system:
- stale leaders and split-brain behavior
- followers not rehoming correctly after takeover
- live sessions disappearing from `/api/sessions`
- leader-assigned session names not matching what the session itself sees locally

The naming problem is now clear enough to document properly before more implementation work.

Today, the system is too close to this anti-pattern:
- followers can have a local idea of their name
- the leader can separately invent or reassign a different name
- reconnects and takeovers can cause churn

That is not a stable architecture for a mixed-version, long-running environment.

## Ideal Architecture
The ideal architecture is **leader-authoritative display-name leasing with follower proposals**.

### Identity model
Every session should carry three different identities:

1. `stable_session_id`
- long-lived identity for the logical session family when available
- should survive reconnects and compatible restarts
- used for continuity and takeover reconciliation

2. `runtime_session_id`
- unique identity for the currently running process/runtime
- used to identify the actual active lease owner

3. `display_name`
- human-friendly puppet name shown in the web console
- must be treated as a leased UI label, not the true identity

### 1. The active leader is the only authority that can grant a display name
Followers may propose a preferred name, but only the leader can finally allocate or confirm it.

Why:
- multiple live processes can race
- mixed-version systems can reconnect unpredictably
- there must be one conflict resolver

### 2. Sessions do not get a brand-new name on every reconnect
Registration must be idempotent.

A reconnect should mean:
- renew existing lease if the same runtime is still the same active session
- resume prior lease when safe
- allocate a new name only when this is truly a new runtime that cannot safely inherit the old lease

### 3. A live runtime must have at most one active display name
At any given time:
- one `runtime_session_id` -> one active `display_name`
- one active `display_name` -> one runtime owner

### 4. The leader must preserve names across takeover whenever possible
If leadership changes:
- the new leader imports the existing session registry
- sessions should keep the same `display_name` when identity matches and the lease is still valid
- a takeover should not casually rename live sessions

### 5. Followers must use the leader-assigned name once one exists
A follower may have a provisional local preference before it reaches a leader.
Once the leader responds with the authoritative assignment, that assignment wins.

### Session startup
1. Session starts.
2. If no leader exists, it becomes leader.
3. If it is leader, it allocates its own display name through the same authoritative allocation path used for all sessions.
4. If a leader exists, the follower sends:
- `stable_session_id`
- `runtime_session_id`
- optional `preferred_display_name`
- optional `last_assigned_display_name`
- any needed protocol/version metadata
5. The leader responds with the authoritative `display_name` and lease state.

### Session heartbeat
Heartbeat should renew the lease and include enough identity to make the operation idempotent.
A heartbeat must not silently create a second name for the same runtime.

### Session reconnect
On reconnect, the leader should resolve in this order:
1. existing active lease for `runtime_session_id`
2. resumable recent lease for `stable_session_id`
3. allocation of a new display name

### Session shutdown or expiry
When a session cleanly ends or its lease expires:
- the display name returns to the pool
- optional cooldown can remain to reduce churn
- the lease record should be clearly closed so the name cannot remain ghost-owned forever

## Pool / Allocation Expectations
The name pool should behave like a central leased resource:
- the leader tracks which names are free, leased, cooling down, or released
- followers should not independently finalize global assignment
- a follower can suggest a preferred name, but the leader grants or rejects it

This gives us the concurrency safety of a central allocator while still letting a session keep a stable self-identity preference.

### Temporary connectivity loss
If a follower temporarily loses connectivity:
- it should keep its current local assigned name as provisional UI state
- it should not allocate itself a new global name
- when connectivity resumes, it should renew/resume the old lease if possible

### Same session reconnecting
A reconnecting session should not be renamed unless the old lease is truly unavailable or invalid.

### Two live runtimes for one stable session
If two runtimes for the same `stable_session_id` are alive at once:
- they must not share one display name concurrently
- the leader must choose one active lease owner per runtime
- policy should be explicit: newest runtime wins, or the currently active lease holder wins until expiry/displacement

### Split-brain leadership
If two processes temporarily believe they are leaders:
- once convergence happens, the surviving leader becomes authoritative
- follower sessions must re-register against that leader
- names should be reconciled by lease ownership, not randomly reassigned

### Old clients / mixed versions
Older sessions may not send the full ideal identity payload.
The leader should have a compatibility path, but the compatibility path should still avoid inventing unnecessary replacement names when a better identity match is available.

## Implementation Guidance
The next implementation should avoid scattering name logic across multiple independent paths.

There should be one authoritative flow that conceptually behaves like:
- `registerOrResumeSessionLease(...)`

That flow should:
1. resolve the leader/session authority
2. match by `runtime_session_id`
3. else match by resumable `stable_session_id`
4. else attempt preferred-name allocation
5. else allocate the next available name
6. return the authoritative assignment for storage, heartbeat, logs, metrics, and UI

### Things we should explicitly avoid
- assigning a new name on every registration
- assigning a new name on heartbeat
- separate naming algorithms in multiple components that can drift
- leader-side fallback that invents names before checking existing identity/lease state
- treating `display_name` as the true identity

## Acceptance Criteria
We should consider this architecture implemented correctly when all of these are true:
- a session sees the same display name locally and in the web console
- reconnecting a session does not usually rename it
- leader takeover does not rename healthy live sessions unnecessarily
- one live runtime never appears with two active names
- two live sessions never share the same active display name
- mixed-version environments converge to a single authoritative session registry
- temporary connectivity loss does not cause name churn
- the leader can reconcile old-version clients without inventing obviously wrong replacements when better identity data exists

## Notes
The right model here is not “nobody assigns names to anyone else.”
The right model is:
- sessions may propose
- the leader grants
- the lease is renewed/resumed idempotently

That is the safe architecture for a concurrent, long-running, mixed-version environment.

mickdarling:
## Summary
- add leader-authoritative session lease registration/resume flow for runtime and stable session identities
- let followers keep a local preferred name until the leader returns an authoritative lease, then reuse that assignment for heartbeat and forwarding
- add regression coverage for lease responses, stable-session resume, and leader-assigned name reuse across forwarding

Closes #2111

## Verification
- npm test -- --runInBand tests/unit/web/console/sessionRegistry.test.ts tests/unit/web/console/console-failure-modes.test.ts tests/unit/web/console/ghostSession.test.ts tests/unit/web/console/UnifiedConsole.test.ts tests/unit/web/console/SessionNames.test.ts
- npx eslint src/web/console/IngestRoutes.ts src/web/console/LeaderForwardingSink.ts src/web/console/UnifiedConsole.ts tests/unit/web/console/sessionRegistry.test.ts tests/unit/web/console/console-failure-modes.test.ts tests/unit/web/console/ghostSession.test.ts tests/unit/web/console/UnifiedConsole.test.ts tests/unit/web/console/SessionNames.test.ts
- npm run build

### DIFF
--- a/src/web/console/IngestRoutes.ts
+++ b/src/web/console/IngestRoutes.ts
@@ -34,6 +34,7 @@ import {
   CONSOLE_PROTOCOL_VERSION,
   LEGACY_CONSOLE_PROTOCOL_VERSION,
 } from './LeaderElection.js';
+import type { ConsoleTokenEntry } from './consoleToken.js';
 
 /** Maximum payload size for ingestion requests */
 const MAX_PAYLOAD_SIZE = '1mb';
@@ -209,6 +210,10 @@ interface SessionLeaseResolution {
   resolution: 'runtime-renewal' | 'stable-resume' | 'new-allocation';
 }
 
+interface IngestRouteLocals {
+  tokenEntry?: ConsoleTokenEntry;
+}
+
 interface SessionDisplayNameHints {
   displayName?: string;
   preferredDisplayName?: string;
@@ -269,6 +274,8 @@ export function createIngestRoutes(broadcasts: IngestBroadcasts): IngestRoutesRe
   // (every 10s) carries the PID — we SIGTERM immediately and move to killedSessions.
   const pendingKills = new Set<string>();
   const importedSessionGraceUntil = new Map<string, number>();
+  // Bounded observability counters for lease-flow debugging. This object keeps a
+  // fixed set of numeric totals and does not grow with session count.
   const metadataSyncCounters = {
     displayNameAdoptions: 0,
     stableSessionBindings: 0,
@@ -643,7 +650,7 @@ export function createIngestRoutes(broadcasts: IngestBroadcasts): IngestRoutesRe
   /**
    * POST /api/ingest/session — Session lifecycle events.
    */
-  router.post('/api/ingest/session', (req: Request, res: Response) => {
+  router.post('/api/ingest/session', (req: Request, res: Response<unknown, IngestRouteLocals>) => {
     const payload = req.body as SessionEventPayload;
     if (!payload?.sessionId || !payload.event) {
       const received = payload ? Object.keys(payload) : [];
@@ -672,7 +679,7 @@ export function createIngestRoutes(broadcasts: IngestBroadcasts): IngestRoutesRe
           lastAssignedDisplayName: payloadLastAssignedDisplayName,
           pid: payload.pid,
           startedAt: payload.startedAt || now,
-          authenticated: Boolean((res as any).locals?.tokenEntry),
+          authenticated: Boolean(res.locals.tokenEntry),
           kind: 'mcp',
           isLeader: false,
           serverVersion: payload.serverVersion,
--- a/src/web/console/IngestRoutes.ts
+++ b/src/web/console/IngestRoutes.ts
@@ -209,6 +209,12 @@ interface SessionLeaseResolution {
   resolution: 'runtime-renewal' | 'stable-resume' | 'new-allocation';
 }
 
+interface SessionDisplayNameHints {
+  displayName?: string;
+  preferredDisplayName?: string;
+  lastAssignedDisplayName?: string;
+}
+
 /** Normalize a string via UnicodeValidator (DMCP-SEC-004) */
 function normalizeInput(s: string): string {
   return UnicodeValidator.normalize(s).normalizedContent;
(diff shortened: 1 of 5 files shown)


mickdarling:
@claude please review the latest commits specifically in regards to any of the issues, recommendations, or other things that needed addressing from previous reviews.

mickdarling:
@claude now please review the latest commit with regards to any of the issues that were brought up in previous rounds. And is this ready to go or does it need more work?

PR #2152, opened 2026-04-21
Emit valid fail-open JSON for hook adapters

mickdarling:
## Summary
- emit explicit allow JSON for shared hook fail-open paths instead of exiting with empty stdout
- do the same for the VS Code wrapper so startup races and transient outages stay JSON-valid
- update the hook contract doc and regression tests to match

## Why
Codex was still hitting  during restart/startup races because the shared hook bridge had early  branches that emitted no JSON at all when the permission server port file was missing, the authority mode was off, the response was malformed, or retries were exhausted.

## Testing
- npm run test:integration -- --runInBand tests/integration/hooks/permission-hook-docker.test.ts
- bash -n scripts/pretooluse-dollhouse.sh scripts/pretooluse-codex.sh scripts/pretooluse-cursor.sh scripts/pretooluse-gemini.sh scripts/pretooluse-vscode.sh

### DIFF
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -49,6 +49,14 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Install ShellCheck
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y shellcheck
+
+      - name: Lint shell scripts
+        run: npm run lint:shell
+
       - name: Build project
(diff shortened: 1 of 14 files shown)


mickdarling:
@claude please review the latest commits specifically in regards to anything outstanding that was called for in earlier reviews. Look at the entire code that exists now. And tell me if it is production ready, you are checking codex's work at this point.

PR #2154, opened 2026-04-21
Fix flaky permission hook integration test on Linux CI

mickdarling:
## Summary
- isolate the port-discovery permission hook test to its own temporary hook home
- stop relying on the shared ~/.dollhouse/run path for that assertion
- stabilize the post-merge Cross-Platform Simple failure on main

## Testing
- npm test -- --runInBand tests/unit/di/permissionServerIntegration.test.ts -t "hook script should discover server via port file and get a response" (run 5 times)


mickdarling:
Superseded by #2155 so the fix follows the hotfix branch naming convention.

PR #2171, opened 2026-04-23
Improve Codex permission audit visibility

### LINKED ISSUE #2170: Codex permission audit loses session identity and records allowed activity as unknown
## Summary

Codex-originated permission activity is currently hard to trust and hard to interpret in DollhouseMCP.

Two related problems are happening at once:

1. The permission audit/dashboard is dropping the identity fields that would let us distinguish normal visible Codex activity from hidden/background suggestion activity.
2. Allowed Codex activity is being recorded as `unknown` instead of `allow` because the Codex hook path intentionally emits empty output on allow, while the audit tracker only looks for top-level `decision` / `behavior` fields.

This is especially confusing when hidden Codex background suggestion threads appear as new sessions and run tool calls that were not typed in the visible conversation.

## User-visible symptoms

Observed on **April 23, 2026** while using Codex Desktop with DollhouseMCP as the permission backend.

- New Dollhouse sessions appeared with no obvious user action behind them.
- The dashboard/audit feed did not clearly identify them as Codex background work.
- A hidden Codex thread ran commands such as `pwd`, `git status`, `git log`, `rg --files`, `ls`, and `echo hi`.
- The permission audit feed showed allowed Codex activity as `unknown` rather than `allow`.
- We still do not have `session_id` in the permission audit feed, which makes tracing this activity much harder.

This is spooky UX. Even if the underlying behavior is legitimate, the current presentation makes it look like unexplained activity.

## Concrete example that triggered this

We traced one hidden Codex thread/session to:

- `session_id`: `019dba80-848a-7793-96a2-e2396fa72cb3`
- origin: internal Codex Desktop background suggestion flow
- commands included: `pwd`, `git rev-parse --show-toplevel`, `git status --short --branch`, `git branch --all --verbose --no-abbrev`, `git log --date=iso --decorate --oneline -12`, `rg --files`, `ls`, `/bin/pwd`, `pwd`, `echo hi`

The raw hook diagnostics clearly identified this as Codex activity, but the main permission dashboard did not surface enough information to make that obvious.

### 1. The hook already knows more than the dashboard shows

`~/.dollhouse/hooks/pretooluse-dollhouse.sh` already forwards `session_id` in the payload sent to `/api/evaluate_permission`:

```bash
PAYLOAD=$(jq -cn \
  --arg tool_name "$TOOL_NAME" \
  --arg platform "$HOOK_PLATFORM" \
  --arg session_id "${DOLLHOUSE_SESSION_ID:-}" \
  --argjson input "$TOOL_INPUT" \
  '{
    tool_name: $tool_name,
    input: $input,
    platform: $platform
  } + (if ($session_id | length) > 0 then { session_id: $session_id } else {} end)')
```

The hook also has access to the full raw Codex hook input on stdin, which in practice includes fields like:

- `session_id`
- `turn_id`
- `tool_use_id`
- `transcript_path`
- `cwd`
- `model`

So the raw signal exists.

### 2. The web route drops that identity context

`src/web/routes/permissionRoutes.ts` currently narrows the request body to only:

- `tool_name`
- `input`
- `platform`

It does **not** carry through `session_id` (or any other hook metadata) into the tracked `recentDecisions` feed.

`PermissionDecision` currently only stores:

- `id`
- `timestamp`
- `tool_name`
- `command?`
- `decision`
- `reason?`

So the permission audit stream loses the fields that would let us identify which session/client/tool-call produced an entry.

### 3. The dashboard is aggregate-only and not identity-aware

`src/web/public/permissions.js` renders the live feed with only:

- time
- decision
- tool/command
- reason

It does not show:

- `session_id`
- `platform`
- client/source label
- `turn_id`
- `tool_use_id`
- whether the activity is likely foreground vs background

That makes the "All Sessions Live Decision Feed" hard to interpret when hidden/background Codex activity is present.

### 4. Codex allow activity becomes `unknown`

`src/web/routes/permissionRoutes.ts` tracks decisions by extracting only top-level `decision` or `behavior` fields.

That works for Claude Code style responses, but Codex uses a different hook shape.

`evaluate_permission` documents Codex responses as:

```json
{ "hookSpecificOutput": { "permissionDecision": "allow|deny", "reason?": "..." } }
```

However, the audit tracker is not reading `hookSpecificOutput.permissionDecision`.

### 5. The Codex hook intentionally emits empty output on allow

This part is important: the client-facing Codex hook behavior is fine.

In `~/.dollhouse/hooks/pretooluse-dollhouse.sh`:

- `build_allow_response()` returns an empty string for `codex`
- `normalize_response()` for `codex` intentionally turns allow-like responses into empty output and only emits structured output for deny

So for Codex, **empty output is the expected allow path**.

That is okay for the hook/client contract.

But upstream in Dollhouse's own audit/decision tracking, we still need to explicitly record that we decided `allow`.

Right now, because the audit path is effectively trying to infer the decision from the client-facing hook response shape, allowed Codex actions are ending up as `unknown` instead of `allow`.

## Why this matters

This is not just a telemetry nicety.

When hidden/background suggestion activity appears, we need the dashboard/audit trail to make it clear that:

- this came from Codex
- which session it came from
- whether it was a visible foreground session or likely background/hidden work
- whether Dollhouse allowed or denied it

Without that, legitimate background behavior looks like unexplained agent activity.

### Minimum

- Preserve `session_id` all the way through the permission audit path.
- Record `platform` in `recentDecisions`.
- Record the actual internal decision for Codex allow/deny instead of inferring only from the client-facing hook output.
- Teach the tracker to understand Codex response shape (`hookSpecificOutput.permissionDecision`).

### Better

- Forward and store additional hook metadata when present:
  - `turn_id`
  - `tool_use_id`
  - `transcript_path`
- Surface those in the dashboard and/or detailed audit view.
- Add a client/source badge such as `Codex`, `Claude Code`, `VS Code`, etc.
- Add a `foreground` / `background` / `unknown` activity classification or similar UX label.

### Codex-specific hinting

For Codex, a useful first-pass heuristic for "background suggestion" appears to be some combination of:

- platform = `codex`
- missing or null transcript/thread metadata
- missing visible thread/app identifiers
- hidden/internal suggestion payloads

Even a simple badge like `Possible Background Suggestion Activity` would be much better than unlabeled aggregate rows.

## Acceptance criteria

- Codex-originated permission events no longer appear as `unknown` when Dollhouse decided `allow`.
- `recentDecisions` includes `session_id` and `platform` at minimum.
- The permission dashboard shows enough session/client context to identify where an entry came from.
- Hidden/background Codex activity is not silently confounded with normal visible session activity.
- The audit path records Dollhouse's decision independently from whether the Codex hook emits empty output to the client.

## Relevant code paths

- `src/web/routes/permissionRoutes.ts`
- `src/web/public/permissions.js`
- `src/handlers/mcp-aql/evaluatePermission.ts`
- `src/handlers/mcp-aql/OperationSchema.ts`
- `~/.dollhouse/hooks/pretooluse-dollhouse.sh`

## Notes

There is also a broader telemetry gap: `src/telemetry/clientDetector.ts` does not currently include Codex as a recognized MCP client. That may be related to some of the missing client labeling, although the more immediate bug here is the permission audit path dropping metadata it already has.

mickdarling:
## Summary
- record Codex allow decisions correctly in the permission audit feed instead of showing them as unknown
- preserve Codex hook metadata like session, turn, tool-use, transcript, cwd, and model through the audit path
- surface platform/session context in the compact permissions feed and expanded audit modal

## Testing
- npm test -- --runInBand tests/unit/web/permissionRoutes.test.ts
- npm test -- --runInBand tests/unit/web/console-ui-regressions.test.ts
- npm run lint:shell
- npx eslint src/web/routes/permissionRoutes.ts src/web/public/permissions.js tests/unit/web/permissionRoutes.test.ts tests/unit/di/permissionServerIntegration.test.ts tests/unit/web/console-ui-regressions.test.ts
- npm run build

Closes #2170


mickdarling:
@claude please review any issues that came up in reviews earlier

PR #2182, opened 2026-05-01
[codex] Remove arbitrary description length cap

### LINKED ISSUE #2181: edit_element accepts >500-char description and writes an unloadable element (roundtrip validation break)
## Summary

`edit_element` (and adjacent write paths) accept a `description` value longer than `MAX_DESCRIPTION_LENGTH` (currently 500) and persist the full overlength string to disk. On the next load, the same value is rejected with `Invalid value for field 'description'`, so the element becomes **invisible to `list_elements`, `search_elements`, and `activate_element`** until a human edits the YAML manually. The on-disk file lies outside what the loader will accept — the write side and read side disagree.

This bit a real session: a `dollhouse-yc-champion` persona was edited to v1.0.1 with a 535-character description, the edit reported success, and the persona then disappeared from the portfolio for the rest of the session. Recovering required hand-trimming the frontmatter (which violates the "always use MCP tools, never edit element files directly" guidance).

## Repro

1. `mcp_aql_update edit_element` on any persona/skill/etc., with `input.description` set to a 501+ character string.
2. Inspect the file in `~/.dollhouse/portfolio/<type>/<name>.md` — full overlength description is on disk.
3. Restart the server (or wait for cache invalidation) and call `get_element` / `activate_element` / `list_elements`.
4. The element is reported as invalid and excluded from listings; activation fails with `Persona not found`.
5. `get_element` returns: `❌ <Type> Invalid — exists on disk but failed to load: Invalid value for field 'description'`.

## Root cause (as I read the code in `@dollhousemcp/mcp-server@2.0.32`)

- `MAX_DESCRIPTION_LENGTH: 500` in `dist/security/constants.js:21`
- `validateAndSanitizeInput` calls `sanitizeInput(input, maxLength)` (`dist/services/validation/ValidationService.js`), which is a **silent truncator** — overlength input is shortened, not rejected
- `editElement.js:144` passes the description through that pipeline with `maxLength: 500`
- But the persisted YAML is the **untruncated original**, not the sanitized result — so the file holds a value the loader will later reject
- `GenericElementValidator.validateDescription` (load path) runs the same `validateAndSanitizeInput` and now **does** fail because the regex+length combination produces an `errors` entry instead of silent truncation

So the bug isn't the regex (`SAFE_DESCRIPTION` is permissive) — it's that **write accepts what read rejects**, and there is no symmetry guarantee.

## Proposed fixes

**Required (the actual bug):**
1. **Make write-side validation hard-fail on overlength descriptions** with a clear error (\"description exceeds 500 characters; got N\"), instead of silently truncating-but-persisting-original. The roundtrip MUST be: anything we accept on write must load on read.
2. **Or: persist the sanitized value** (truncate-and-save) and warn the caller that truncation occurred. Either is a roundtrip guarantee; the current behavior is neither.

**Desirable (the feature ask, raised by @mickdarling):**
3. **Make `MAX_DESCRIPTION_LENGTH` configurable.** 500 is restrictive for substantive personas/agents that genuinely need 600-1500 chars of self-description. This is not a YAML constraint — YAML handles arbitrary string length — it is a sanitization constant whose rationale (DoS? display?) is undocumented. Suggested surface:
   - A field on `dollhouse_config` (e.g. `validation.maxDescriptionLength`, default 500, capped at some hard ceiling like 5000)
   - Or env var `DOLLHOUSE_MAX_DESCRIPTION_LENGTH`
   - Whatever value is configured must be applied **identically on write and on read** (the bug above must already be fixed first)
4. **Surface the limit in error messages and in `introspect operations`** so callers know what to target without reading source.

## Suggested test coverage

- Roundtrip test: for every element type, write an element via the public CRUD operations with `description` lengths at `[1, MAX-1, MAX, MAX+1, MAX*2]` and assert `get_element` succeeds iff the write succeeded.
- Negative test: an overlength description on `edit_element` returns a structured error and **does not modify the file on disk**.

## Severity

Practical impact: an element gets written, disappears from the portfolio, and a user has to know to edit YAML by hand to recover it. This is exactly the failure mode "always use MCP tools, never file-edit elements" is supposed to prevent — yet here the only recovery is a file edit.

## Environment

- `@dollhousemcp/mcp-server@2.0.32` (npm)
- Encountered against a persona whose description was edited via `edit_element` to remove training-corpus deadline anxiety; the resulting 535-char description was persisted but unloadable.

mickdarling:
## Summary

Fixes #2181.

Removes the historical 500-character hard cap from Dollhouse element descriptions and description-like metadata fields. Descriptions are no longer blocked by the short metadata-field limit, but they remain bounded by the existing YAML/frontmatter and file safety limits so create/edit cannot write unloadable elements.

This PR is built from and targets `develop`. It stays focused on length handling only; content-safety pattern checks such as XSS/prompt-injection detection are unchanged.

## What changed

- Removed `MAX_DESCRIPTION_LENGTH` and replaced description-specific validation with the existing YAML/frontmatter safety boundary for metadata descriptions.
- Updated secure YAML parsing, content metadata validation, generic element validation, CRUD create/edit handlers, and element constructors/managers.
- Added recursive create/edit validation for nested `description` fields, including `metadata.description`, template variables/examples, skill parameters, and agent goal parameters.
- Updated tests and docs so >500 character descriptions are valid, while descriptions beyond the frontmatter safety limit are rejected before save.
- Fixed the stale `docs/architecture/ensemble-system.md` reference called out by Claude.

## Validation

- `npm test -- --runInBand tests/unit/handlers/element-crud/createElement.helper.test.ts tests/unit/handlers/element-crud/editElement.helper.test.ts tests/security/inputLengthValidation.test.ts tests/unit/services/SerializationService.test.ts tests/unit/services/validation/GenericElementValidator.test.ts tests/unit/base/ElementValidation.test.ts --modulePathIgnorePatterns='<rootDir>/.worktrees'`
- `npm test -- --runInBand tests/unit/elements/skills/SkillManager.test.ts tests/unit/elements/templates/TemplateManager.test.ts tests/unit/elements/agents/AgentManager.test.ts tests/unit/elements/templates/Template.test.ts tests/unit/elements/agents/Agent.test.ts tests/unit/elements/memories/Memory.test.ts tests/unit/elements/ensembles/Ensemble.test.ts --modulePathIgnorePatterns='<rootDir>/.worktrees'`
- `npm run test:integration -- --runInBand tests/integration/mcp-aql/update.test.ts --modulePathIgnorePatterns='<rootDir>/.worktrees'`
- `npm run test:integration -- --runInBand tests/integration/crud/ElementCRUD.test.ts --modulePathIgnorePatterns='<rootDir>/.worktrees'`
- `npx tsc --noEmit --pretty false`
- `git diff --check`

### DIFF
--- a/docs/architecture/element-manager-consolidation.md
+++ b/docs/architecture/element-manager-consolidation.md
@@ -720,7 +720,7 @@ export abstract class BaseElementManager<T extends IElement> {
     // Sanitize inputs
     const sanitizedName = sanitizeInput(metadata.name, 100);
     const sanitizedDesc = metadata.description ?
-      sanitizeInput(metadata.description, SECURITY_LIMITS.MAX_CONTENT_LENGTH) : '';
+      sanitizeInput(metadata.description, SECURITY_LIMITS.MAX_YAML_LENGTH) : '';
     const sanitizedContent = content ?
       sanitizeInput(content, 50000) : '';
 
@@ -1094,7 +1094,7 @@ const skillConfig: ElementTypeConfig = {
   },
   validation: {
     namePattern: /^[a-zA-Z0-9-]+$/,
-    descriptionMaxLength: SECURITY_LIMITS.MAX_CONTENT_LENGTH,
+    descriptionMaxLength: SECURITY_LIMITS.MAX_YAML_LENGTH,
     contentMaxLength: 50000,
     customValidation: (element) => {
       const skill = element as Skill;
@@ -1126,7 +1126,7 @@ const agentConfig: ElementTypeConfig = {
   },
   validation: {
     namePattern: /^[a-zA-Z0-9_-]+$/,
-    descriptionMaxLength: SECURITY_LIMITS.MAX_CONTENT_LENGTH,
+    descriptionMaxLength: SECURITY_LIMITS.MAX_YAML_LENGTH,
     contentMaxLength: 50000
   },
   fileExtension: '.md',
--- a/docs/architecture/element-manager-consolidation.md
+++ b/docs/architecture/element-manager-consolidation.md
@@ -720,7 +720,7 @@ export abstract class BaseElementManager<T extends IElement> {
     // Sanitize inputs
     const sanitizedName = sanitizeInput(metadata.name, 100);
     const sanitizedDesc = metadata.description ?
-      sanitizeInput(metadata.description, 500) : '';
+      sanitizeInput(metadata.description, SECURITY_LIMITS.MAX_CONTENT_LENGTH) : '';
     const sanitizedContent = content ?
       sanitizeInput(content, 50000) : '';
 
@@ -1094,7 +1094,7 @@ const skillConfig: ElementTypeConfig = {
   },
   validation: {
     namePattern: /^[a-zA-Z0-9-]+$/,
-    descriptionMaxLength: 500,
+    descriptionMaxLength: SECURITY_LIMITS.MAX_CONTENT_LENGTH,
     contentMaxLength: 50000,
     customValidation: (element) => {
       const skill = element as Skill;
@@ -1126,7 +1126,7 @@ const agentConfig: ElementTypeConfig = {
   },
   validation: {
     namePattern: /^[a-zA-Z0-9_-]+$/,
-    descriptionMaxLength: 500,
+    descriptionMaxLength: SECURITY_LIMITS.MAX_CONTENT_LENGTH,
     contentMaxLength: 50000
   },
   fileExtension: '.md',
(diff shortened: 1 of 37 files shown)


mickdarling:
@claude please do a fresh review of this PR after the latest Sonar cleanup. Please specifically verify the issues and suggestions from the previous reviews:

- The old docs concern for `docs/architecture/ensemble-system.md` should now be resolved; it should no longer claim a 500-character description cap.
- Description validation should no longer use the arbitrary 500-character cap, and should be bounded by the YAML/frontmatter safety limit for metadata descriptions rather than the body content limit.
- Nested `description` fields in create/edit payloads should be handled consistently, including template variables/examples, agent goal parameters, skill parameters, memory fields, and ensemble fields.
- The new tests should cover valid descriptions over 500 characters and rejection only when descriptions exceed the YAML/frontmatter safety boundary.
- The three SonarCloud maintainability findings from the previous analysis should be addressed by the latest commit.

Please call out any remaining correctness or test gaps, especially around long Dollhouse element descriptions.

PR #2196, opened 2026-05-15
fix: pin patched lodash, path-to-regexp, picomatch, brace-expansion

mickdarling:
## Summary

Clears npm-audit findings on transitive dependencies via targeted `overrides`. Companion to #2195 (which handled the Dependabot-tracked alerts); this covers the npm-audit-only findings.

Uses **version-range-scoped override keys** to patch within each major line. This keeps the `package-lock.json` diff surgical (65 lines) instead of the ~1900-line churn a global override + `npm dedupe` would produce.

| Pkg | Was | Now | Scope | Issue |
|---|---|---|---|---|
| lodash | 4.17.23 | 4.18.1 | runtime (archiver→archiver-utils) | code injection via `_.template`, prototype pollution in `_.unset`/`_.omit` |
| path-to-regexp | 8.2.0 | 8.4.2 | runtime (express→router) | ReDoS — sequential optional groups / multiple wildcards |
| picomatch | 2.3.1 / 4.0.3 | 2.3.2 / 4.0.4 | devDep (jest) | method injection, ReDoS via extglob |
| brace-expansion | 1.1.12 / 2.0.2 | 1.1.13 / 2.0.3 | devDep (eslint→minimatch) | zero-step sequence DoS |

`path-to-regexp` is scoped under `express` so `serve-handler`'s 3.x (not vulnerable to these CVEs) is untouched. picomatch/brace-expansion are patched within-major to avoid forcing major upgrades on consumers.

## Known residual (intentional, low risk)

`brace-expansion` in the **`packages/safety` workspace member** still resolves to 1.1.12. npm does not apply root `overrides` to workspace members that run their own `npm install` (the `build:safety` script). This is:

- **build/lint-only** — reaches brace-expansion via `eslint → minimatch`
- **never shipped** — `@dollhousemcp/safety` publishes only `dist/`
- **never in the runtime path**

The proper fix is workspace build hygiene (drop the redundant nested `npm install`, rely on workspace hoisting), which is a build-pipeline change deliberately kept out of this security PR. Tracked for follow-up.

## Out of scope

`install-mcp → giget → tar` (3 findings) — handled separately (Group B). Reachability analysis shows DollhouseMCP only invokes install-mcp in command-mode (config write), never the giget tarball-extraction path the tar CVEs require.

## Test plan

- [x] `npm ci` reproduces fixes deterministically (verified the lockfile, not just a dirty install)
- [x] `npm run build` passes
- [x] 10367/10377 unit tests pass (1 pre-existing timing flake in `permissionRoutes`, unrelated)
- [ ] CI on this PR

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1204,7 +1204,9 @@
       }
     },
     "node_modules/@eslint/config-array/node_modules/brace-expansion": {
-      "version": "1.1.12",
+      "version": "1.1.13",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.13.tgz",
+      "integrity": "sha512-9ZLprWS6EENmhEOpjCYW2c8VkmOvckIJZfkr7rBW6dObmfgJ/L1GpSYW5Hpo9lDz4D1+n0Ckz8rU7FwHDQiG/w==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -1270,7 +1272,9 @@
       }
     },
     "node_modules/@eslint/eslintrc/node_modules/brace-expansion": {
-      "version": "1.1.12",
+      "version": "1.1.13",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.13.tgz",
+      "integrity": "sha512-9ZLprWS6EENmhEOpjCYW2c8VkmOvckIJZfkr7rBW6dObmfgJ/L1GpSYW5Hpo9lDz4D1+n0Ckz8rU7FwHDQiG/w==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


mickdarling:
Re: Codex review feedback on this PR (npm `overrides` are ignored when this package is installed as a dependency, so the fix doesn't reach end users via npx/npm i) — **addressed in #2198**.

#2198 commits `npm-shrinkwrap.json` (the one lockfile npm publishes *and* honors for consumers), generated from the fully-patched develop tree, shipped via the `files` allowlist. Dual-file design keeps `package-lock.json` untouched (zero changes to production code/Docker/CI). A `prepublishOnly` guard aborts publish on a stale shrinkwrap, and a new `lockfile-sync.yml` CI check fails any PR where the two lockfiles diverge.

This PR's overrides are still correct and necessary (they fix the repo's own dev/CI tree); #2198 is what propagates them to published consumers.

PR #2197, opened 2026-05-15
fix: scope giget's tar to 7.5.15 (clears install-mcp/giget/tar CVEs)

mickdarling:
## Summary

Stopgap for the `install-mcp → giget → tar` advisory chain (3 high npm-audit findings). All three are flagged **only "via tar"**, so a single scoped tar pin clears the whole chain.

```json
"overrides": { "giget": { "tar": "7.5.15" } }
```

Scoped under `giget` — narrow blast radius (one `tar` in the tree, used solely by giget). `tar@7.5.15` fixes the hardlink/symlink path-traversal + drive-relative-linkpath CVEs (vulnerable `<=7.5.10`).

## Reachability (why this is low real-risk, but still worth fixing)

DollhouseMCP spawns `install-mcp` in **command-mode only** (`setupRoutes.ts` → `execFile install-mcp @dollhousemcp/mcp-server@<tag> --client X --name dollhousemcp --yes`), which writes a client config via `fs.writeFileSync`. The tar CVEs require giget's **tarball-extraction** path (URL/template fetch mode), which DollhouseMCP never invokes.

Caveat handled: `giget@1.2.5` does `require('tar')` eagerly at module load, so the module must stay loadable under tar@7. Verified — see test plan.

## Upstream plan (the real fix)

`install-mcp` is upstream `supermemoryai/install-mcp` (MIT). The durable fix is bumping its `giget@^1` → `giget@^3` (modern tar) upstream, then dropping this override once a fixed install-mcp publishes. This PR is the interim.

## Test plan

- [x] `npm ci` reproduces deterministically; lockfile diff is a net reduction (old tar@6 subtree removed)
- [x] `install-mcp --help` / `--version` run cleanly under tar@7.5.15
- [x] `giget` module loads under tar@7.5.15 (eager `require('tar')` succeeds)
- [x] `setupRoutes` + `setupRoutes.nvm` suites: 332/332 pass (the install-mcp config flow)
- [x] Full unit suite: 10367/10377 (1 pre-existing `permissionRoutes` timing flake, unrelated)
- [ ] CI on this PR

## Relationship to other PRs

- #2195 — Dependabot-tracked alerts (fast-uri, hono, ip-address, uuid)
- #2196 — npm-audit-only Group A (lodash, path-to-regexp, picomatch, brace-expansion)
- **This** — npm-audit-only Group B (install-mcp/giget/tar)

Combined, these clear every npm-audit finding except a documented build-only `brace-expansion` residual in the `packages/safety` workspace member.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1465,6 +1465,18 @@
         "node": ">=12"
       }
     },
+    "node_modules/@isaacs/fs-minipass": {
+      "version": "4.0.1",
+      "resolved": "https://registry.npmjs.org/@isaacs/fs-minipass/-/fs-minipass-4.0.1.tgz",
+      "integrity": "sha512-wgm9Ehl2jpeqP3zw/7mo3kRHFp5MEDhqAdwy1fTGkHAwnkGOVsgpvQhL8B5n1qlb01jV3n/bI0ZfZp5lWA1k4w==",
+      "license": "ISC",
+      "dependencies": {
+        "minipass": "^7.0.4"
+      },
+      "engines": {
+        "node": ">=18.0.0"
+      }
+    },
     "node_modules/@istanbuljs/load-nyc-config": {
       "version": "1.1.0",
       "dev": true,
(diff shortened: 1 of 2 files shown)


mickdarling:
Re: Codex review feedback on this PR (npm `overrides` are ignored when this package is installed as a dependency, so the fix doesn't reach end users via npx/npm i) — **addressed in #2198**.

#2198 commits `npm-shrinkwrap.json` (the one lockfile npm publishes *and* honors for consumers), generated from the fully-patched develop tree, shipped via the `files` allowlist. Dual-file design keeps `package-lock.json` untouched (zero changes to production code/Docker/CI). A `prepublishOnly` guard aborts publish on a stale shrinkwrap, and a new `lockfile-sync.yml` CI check fails any PR where the two lockfiles diverge.

This PR's overrides are still correct and necessary (they fix the repo's own dev/CI tree); #2198 is what propagates them to published consumers.

Issue #1629, opened 2026-03-22
refactor: Extract telemetry logic from EnhancedIndexManager


mickdarling:
## Context

Follow-up from #1299 (EnhancedIndexManager refactoring). The main file is down from 2,338 to 1,390 lines after extracting 5 helper modules, but ~100 lines of telemetry logic remain that are orthogonal to core indexing:

- `startTelemetry()`
- `recordTelemetry()`
- `isTelemetryEnabled()`
- `scheduleTelemetryReport()`
- `reportTelemetry()`
- `telemetryMetrics: Map`
- `telemetryTimer: NodeJS.Timeout`

## Proposal

Extract to `IndexTelemetry` or `TelemetryManager` class. Self-contained, doesn't need other EnhancedIndexManager state. Would reduce the file to ~1,290 lines.

## Priority

Low — optional cleanup, no functional impact.

Issue #1608, opened 2026-03-19
refactor: reduce cognitive complexity in 147 flagged functions


mickdarling:
## Summary

SonarCloud flagged 147 functions with cognitive complexity exceeding the threshold of 15 (rule S3776). These were identified during the v2.0.0-rc.1 migration PR (#1607).

## Top Offenders

| File | Line | Complexity | Threshold |
|------|------|-----------|-----------|
| `AgentManager.ts` | 2197 | 69 | 15 |
| `AgentSkillConverter.ts` | 413 | 59 | 15 |
| `AgentManager.ts` | 892 | 49 | 15 |
| `MemorySearchIndex.ts` | 512 | 48 | 15 |
| `CollectionHandler.ts` | 268 | 43 | 15 |
| `AgentSkillConverter.ts` | 625 | 41 | 15 |
| `AgentSkillConverter.ts` | 948 | 35 | 15 |
| `Container.ts` | 946 | 33 | 15 |
| `MemoryManager.ts` | 915 | 30 | 15 |
| `GitHubAuthHandler.ts` | 38 | 29 | 15 |

## Scope

- 147 functions across ~30 source files
- No test files (test complexity is acceptable)
- Concentrated in element managers, converters, and handlers

## Approach

Chip away incrementally rather than one massive refactor:
- Start with the highest complexity functions (>40)
- Extract helper functions, use early returns, simplify branching
- Ensure test coverage before refactoring
- Each PR should target 1-2 files

## Context

These are code smells, not bugs. All flagged functions work correctly and are well-tested. The complexity accumulated during rapid v2 development and is a natural refactoring target for the post-RC1 phase.

Flagged by SonarCloud during PR #1607.

Issue #1179, opened 2025-09-28
🎨 [Maintainability] Reduce cognitive complexity - 108 issues


mickdarling:
## 📋 Issue Category
**Parent Issue**: #1169
**SonarCloud Category**: Maintainability
**Impact**: HIGH
**Estimated Effort**: 8-12 hours (requires careful refactoring)

## 🎯 Scope
Refactor functions with high cognitive complexity (~108 occurrences).

## 📝 Issue Details
- **Rule**: typescript:S3776
- **Description**: Functions should not be too complex
- **Default Threshold**: 15 (cognitive complexity score)

## 🧠 What is Cognitive Complexity?
Counts the mental effort needed to understand code:
- +1 for each: if, else if, else, switch, for, while, do-while, catch
- +1 for each logical operator (&&, ||)
- +1 for nested structures (multiplied by nesting level)

### 1. Extract Methods
```typescript
// Before - High complexity
function processData(data: any) {
  if (data) {
    if (data.type === 'A') {
      // lots of logic
    } else if (data.type === 'B') {
      // lots of logic
    }
  }
}

// After - Lower complexity
function processData(data: any) {
  if (\!data) return;
  
  switch(data.type) {
    case 'A': return processTypeA(data);
    case 'B': return processTypeB(data);
  }
}
```

### 2. Early Returns
```typescript
// Before
function validate(input) {
  if (input) {
    if (input.length > 0) {
      if (input.match(pattern)) {
        return true;
      }
    }
  }
  return false;
}

// After
function validate(input) {
  if (\!input) return false;
  if (input.length === 0) return false;
  if (\!input.match(pattern)) return false;
  return true;
}
```

### 3. Replace Conditionals with Map/Strategy
```typescript
// Before
if (type === 'email') { /* logic */ }
else if (type === 'phone') { /* logic */ }
else if (type === 'address') { /* logic */ }

// After
const handlers = {
  email: handleEmail,
  phone: handlePhone,
  address: handleAddress
};
handlers[type]?.(data);
```

## 🔍 Finding Complex Functions
```bash
# SonarCloud will list them, but you can also use:
npx eslint . --rule 'complexity: ["error", 15]'
```

## ⚠️ Important Notes
- Don't just split code arbitrarily
- Maintain logical cohesion
- Consider readability over metrics
- Add unit tests before refactoring

## ✅ Definition of Done
- [ ] Top 20 most complex functions refactored
[...]

Issue #1169, opened 2025-09-28
🎯 [SonarCloud] Master Tracking Issue - Reduce Technical Debt to Zero


mickdarling:
### Overall Metrics
- **Quality Gate**: ✅ PASSING
- **Reliability**: 289 issues (Grade C)
- **Maintainability**: 2,664 issues (Grade A)
- **Security**: 0 vulnerabilities (Grade A)
- **Security Hotspots**: 214 (need review)
- **Duplications**: 1.0%

## 🎯 Goal
Systematically address all SonarCloud issues to achieve:
- 0 Reliability issues
- 0 Maintainability issues  
- 0 Security Hotspots
- Maintain <1% duplication

### Reliability (289 issues)
- [ ] #___ - Modernize Number parsing methods (180 issues)
- [ ] #___ - Modernize String methods (126 issues)
- [ ] #___ - Fix control character issues (24 issues)
- [ ] #___ - Fix array creation patterns (12 issues)
- [ ] #___ - Remove useless object instantiation (6 issues)
- [ ] #___ - Fix remaining reliability issues (10 issues)

### Maintainability (2,664 issues)
- [ ] #___ - Add node: prefix to imports (495 issues)
- [ ] #___ - Mark readonly properties (175 issues)
- [ ] #___ - Remove unused imports (141 issues)
- [ ] #___ - Reduce cognitive complexity (108 issues)
- [ ] #___ - Fix remaining maintainability issues

### Security Hotspots (214 issues)
- [ ] #___ - Review DOS vulnerabilities (88 issues)
- [ ] #___ - Review authentication hotspots (48 issues)
- [ ] #___ - Review cryptography usage (40 issues)
- [ ] #___ - Review remaining security hotspots (38 issues)

## 📈 Progress Tracking
Will update this issue as sub-issues are completed.

## 🔗 References
- [SonarCloud Project](https://sonarcloud.io/project/overview?id=DollhouseMCP_mcp-server)
- Previous work: PR #1163 (reduced bugs by 82%)

PR #2218, opened 2026-05-25
refactor: reduce cognitive complexity in portfolio/agent managers and split AS

insomnolence:
## Summary

Clears the remaining SonarCloud S3776 cognitive-complexity findings in
the portfolio/agent manager cluster (23 sites across 5 files) and splits
the 1,364-line `EmbeddedAuthorizationServer.ts` into focused per-concern
modules. No public API change; existing tests pass unchanged.

### Cognitive-complexity sweep (23 sites cleared)

- **`src/elements/agents/AgentManager.ts`** — 7 sites including
  `parseMetadata` (was 121), `validateV2FieldsForCreate` (was 69),
  `executeAgent` (was 49), plus `create`, `validateParameters`,
  `detectActivationCycles`, and `serializeElement`. Each split into
  focused private helpers; the parent methods become orchestrators.
- **`src/portfolio/PortfolioSyncManager.ts`** — 7 sites across the sync
  flow (`handleSyncOperation`, `downloadElement`, `uploadElement`,
  `compareVersions`, `bulkDownload`, `bulkUpload`, `calculateSimilarity`).
- **`src/portfolio/EnhancedIndexManager.ts`** — 4 sites including
  `buildIndex`, `loadEnhancedIndexConfig`, `getTriggerMetrics`,
  `searchEnhanced`.
- **`src/portfolio/GitHubPortfolioIndexer.ts`** — 3 sites including
  `getIndex`, `fetchElementTypeContent`, `createGitHubIndexEntry`.
- **`src/portfolio/UnifiedIndexManager.ts`** — 2 sites including
  `calculateDuplicatesCount` (was 44).

### EmbeddedAuthorizationServer split

`src/auth/embedded-as/EmbeddedAuthorizationServer.ts` reduced from 1,364
to 998 lines by extracting into four sibling modules:

- `EmbeddedASBootstrap.ts` (129 LOC) — bootstrap gate, readiness, mode
  detection
- `EmbeddedASTokens.ts` (149 LOC) — `validate`/`issue` and JWT helpers
- `EmbeddedASAdmin.ts` (60 LOC) — admin endpoints
- `EmbeddedASOidcAccount.ts` (52 LOC) — OIDC `findAccount` / `claims`

Public class surface and method signatures preserved.

## Bug-class fixes (caught by expanded local lint)

- `EmbeddedAuthorizationServer.ts` — the `state.provider.callback()(req,
  res)` calls inside the router handler are now awaited so async
  rejections from the OIDC provider reach the surrounding `try/catch`
  and flow to Express's `next(err)`. Without the await, async errors
  during token issuance or adapter operations became unhandled
  rejections.
- `GitHubPortfolioIndexer.ts` frontmatter regexes — narrowed each
  `/^FIELD:\s*(.+)$/m` to `/^FIELD:[ \t]?([^\r\n]+)$/m` so the leading
  whitespace quantifier can't overlap with the value class. Removes the
  polynomial-backtracking shape (regexp/no-super-linear-backtracking).

## Related issues

Related to #1169 (master SonarCloud tracking), #1179 (cognitive
complexity bucket — 108 issues), and #1608 (147 flagged functions);
this PR clears 23 of the cited sites including the highest-complexity
ones in `AgentManager.ts` listed in #1608's top-offenders table.
#1629 (telemetry extraction from `EnhancedIndexManager`) is related
but not addressed — we only verified the `Math.random()` site noted
there is for telemetry sampling, not security.

## Test plan

- [x] TypeScript compile clean
- [x] ESLint with `max-warnings=0`
- [x] Local sonar-mirror lint: all 23 cognitive-complexity sites cleared
- [x] Unit tests pass with no test modifications
- [x] Integration tests pass
- [x] E2E tests pass

### DIFF
--- a/src/auth/embedded-as/EmbeddedASAdmin.ts
+++ b/src/auth/embedded-as/EmbeddedASAdmin.ts
@@ -1,0 +1,60 @@
+import type { RequestHandler } from 'express';
+
+import type { IAuthProvider } from '../IAuthProvider.js';
+import { assertHasRole } from '../assertHasRole.js';
+import { createUnifiedAuthMiddleware } from '../authMiddleware.js';
+import type { IAuthStorageLayer } from './storage/IAuthStorageLayer.js';
+
+export class EmbeddedASAdmin {
+  constructor(
+    private readonly provider: IAuthProvider,
+    private readonly storage: IAuthStorageLayer,
+    private readonly getProtectedResourceMetadataUrl: () => string,
+  ) {}
+
+  createAdminMeHandler(): RequestHandler[] {
+    const validateBearer = createUnifiedAuthMiddleware({
+      provider: this.provider,
+      protectedResourceMetadataUrl: this.getProtectedResourceMetadataUrl(),
+    });
+    const adminGuard = assertHasRole('admin');
+    const handler: RequestHandler = (req, res, next) => {
+      void (async () => {
+        try {
+          const claims = res.locals.authClaims;
+          if (!claims) {
+            next(new Error('admin/me handler reached without authClaims'));
+            return;
+          }
+          const bootstrap = await this.storage.getBootstrapState();
+          const account = await this.storage.getAccount(claims.sub);
+
+          if (!account) {
+            res.status(410).json({
+              error: 'admin account no longer exists',
+              sub: claims.sub,
+            });
+            return;
+          }
+
+          const callerIsBootstrapAdmin = bootstrap.adminSub === claims.sub;
+          res.json({
+            sub: claims.sub,
+            roles: claims.roles ?? [],
+            email: account.email,
+            displayName: account.displayName,
+            bootstrap: {
+              adminMethod: bootstrap.adminMethod,
+              completedAt: bootstrap.completedAt,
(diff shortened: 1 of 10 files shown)


insomnolence:
## On the 8 remaining Sonar S1874 (deprecated) findings

All 8 are reads/writes of `decisionFramework`, `learningEnabled`, and `ruleEngineConfig` on `AgentMetadataV2`. The fields carry `@deprecated v1.x` in the type definition — Sonar's S1874 fires on every access, which is the rule working correctly.

These accesses are the v1.x → v2 migration path. When an agent file written under the v1.x schema is loaded, this code reads the fields to preserve the user's data into the v2 shape; the symmetric serialization code writes them back so a round-trip doesn't drop data. Removing the reads would break every v1.x agent that exists on disk.

We're choosing to leave the S1874 warnings visible rather than refactor or rewrite the migration path. The warnings are correctly flagging the v1.x compatibility code — they're meant as a signal to future authors not to add new uses of these fields, not a request to delete the existing migration logic. When v1.x support is dropped, these reads will go away and the warnings with them.

**Sites included:**
- `AgentManager.ts:398, 400` — normalization reads on the create path
- `AgentManager.ts:2368, 2370, 2383` — read/write pairs on the serialization path

Issue #1965, opened 2026-04-14
Evaluate TypeScript-first web console architecture


mickdarling:
## Summary

The web console is currently implemented as a growing set of raw browser-side JavaScript files. As the console becomes more central to Dollhouse infrastructure and we plan for broader web extensibility, it would be valuable to evaluate moving the console toward a TypeScript-first frontend architecture.

## Motivation

Claude review feedback on recent PRs repeatedly pointed out that the web UI logic would benefit from stronger typing and more structure. We are now building increasingly complex session, permissions, metrics, and audit UX in the console, and the current raw JavaScript approach makes that work harder to maintain and extend safely.

## Explore

- Define what a gradual migration path would look like rather than a full rewrite
- Evaluate TypeScript for existing browser modules even before choosing a framework
- Consider whether a lightweight framework or component model would improve maintainability
- Consider the implications for plugin/extensibility plans for the Dollhouse web server and UI
- Identify which console areas would benefit most from typed models and shared state management
- Evaluate build/test impacts and how to preserve current deployment simplicity

## Desired outcome

A recommendation for whether the console should remain progressively enhanced vanilla JS with TypeScript added, or move toward a fuller TypeScript-based frontend architecture over time.

Issue #1780, opened 2026-04-04
sec: add authentication to web console port 3939


mickdarling:
## Summary

Port 3939 (the DollhouseMCP management console) currently has zero authentication. Any process on `127.0.0.1` can access all endpoints. This needs authentication before the server is used in any multi-session, multi-user, or bridge-connected environment.

## Current state

**Binding**: `127.0.0.1` only (not internet-exposed) — this is the sole security boundary today.

**Unauthenticated endpoints on port 3939:**
- `GET /api/elements` — read all portfolio elements
- `GET /api/logs/stream` — SSE stream of all logs (may contain sensitive operation details)
- `GET /api/metrics/stream` — SSE stream of system metrics
- `GET /api/permissions/status` — current permission policies and recent decisions
- `POST /api/evaluate_permission` — evaluate and approve/deny tool permissions
- `POST /api/ingest/logs` — inject log entries (follower sessions forward here)
- `POST /api/ingest/metrics` — inject metric snapshots
- `POST /api/ingest/session` — register a session
- `POST /api/sessions/:id/kill` — kill a session
- `POST /api/setup/install` — install DollhouseMCP to MCP clients (writes config files)
- `GET /api/sessions` — list all active sessions

**No auth on any of these.** No API keys, no tokens, no session cookies.

## Why this matters

Without authentication, anyone who can reach port 3939 can:
- Inject data into log and metric streams (via ingest endpoints)
- Read all logs including security-sensitive operations
- Approve or deny tool permissions on behalf of the user
- Kill sessions
- Install MCP configurations to client apps

On a single machine with `127.0.0.1` binding, this is limited to local processes. But:
- **Multi-session setups** (multiple bridge sessions) have no isolation on the HTTP layer
- **Docker with port mapping** could accidentally expose 3939
- **Local processes** can interact with the permission system unchecked

### Phase 1: Session token auth (minimum viable)
- Generate a random session token on server startup
- Print it to stderr alongside the console URL
- Require `Authorization: Bearer <token>` on all POST endpoints
- Require token as query param or cookie for SSE streams
- Browser gets the token from the initial page load (embedded in HTML or via a bootstrap endpoint)

### Phase 2: Per-session identity
- Named sessions with individual tokens
- Per-session permission scoping (session A can't kill session B)
- Audit trail: which session approved which permission

### Phase 3: Remote access (if needed)
- TLS on port 3939
- Configurable bind address
- Optional OAuth/OIDC integration

## Priority

[...]

PR #2242, opened 2026-06-05
feat(web-console): /api/v1 BFF console rewrite

insomnolence:
Introduces the `/api/v1` BFF console — a durable, authenticated replacement for the legacy `src/web/console/`. It ships dormant behind an opt-in activation gate, so it has no effect on a running deployment until cutover, and refuses to mount while the legacy console is still enabled.

## Platform kernel & module system
- Route-descriptor model: each route declares method/path/audience, required capability, ownership, elevation tier, privacy class, and idempotency.
- Privacy projectors enforce a field allowlist at the trust boundary (success bodies projected, problem bodies preserved).
- Per-route dependency declaration that fails closed — routes with undeclared deps are refused at activation.
- Idempotency with a claim/complete state machine; periodic store cleanup.

## Durable PostgreSQL backends
- Identity, session, transaction, and idempotency stores.
- Portfolio element store backed by the existing element managers (not envelope-only).
- Session activation-state store and runtime session control-plane substrate.
- HMAC-chained admin audit writer with chain verification on read.
- Security-invalidation processor with a stable per-replica identity and full backlog drain at startup.

## Authentication & elevation
- BFF OAuth + PKCE login with platform-owned cookie issuance.
- TOTP factor storage, enrollment, proof, and disable.
- Graduated admin step-up with administrative ACR issuance and console-only admin roles.
- Sign-in allowlist cut over to the console account allowlist.
- Store-backed JWKS served with rotation overlap.
- BFF security layer: auth, CSRF, crypto, security headers, protected correlation-resolution rate limiting.

## Surfaces
- Self-service: profile/settings, browser session listing and revocation, security factor status.
- Portfolio: read-only listing plus create/update/delete and a sync job queue.
- Account admin: role grant/revoke/replace, invites, disable/enable, credential and OAuth grant revoke-all, with role-tier authority enforcement and a mutation-transaction runner.
- Operator: health, logs, metrics, and configuration read/update.
- Admin audit: read surface with chain verification and SSE export.
- Runtime MCP sessions bridged into the control plane, with list and termination routes; terminations cascade from disable and credential revoke.
- Server-sent-events streaming for session logs, metrics, and executions.

## UI
- Vanilla per-tab modules under `src/web-console/ui` — sessions with request/error counts, admin metrics bridging operational metrics, integrations catalog, and step-up elevation.

## Hosted activation gating
- Production-readiness checks derive activation from the deployment signal and database verification.
- Replacement-readiness CLI and evidence utility; explicit production-adapter metadata required at mount.

References #1965, #1780.


insomnolence:
@claude  Thanks for the review. Three items don't hold up on inspection — each verifiable in a minute.

  1. "Only ~10 web-console test files; no tests for ConsoleSecurityInvalidationProcessor or audit chain."
  The PR adds 79 test files (53 web-console unit tests) under tests/unit/web-console/. The files called out as missing exist:
  services/ConsoleSecurityInvalidationProcessor.test.ts, modules/PostgresAdminAuditQuery.test.ts + ProductionAuditQueries.test.ts (audit chain),
  totp/AdminTotpService.test.ts. The "10" likely came from searching tests/web-console/ instead of tests/unit/web-console/.

  2. "ConsoleAuthentication.ts:46-50 — DB error could bypass auth."
  That snippet is inside a try/catch starting at line 44. next() and req.consoleAuthentication are set only inside the try, after all checks pass; any
  error hits the catch → 503 or next(error), so the request never reaches the handler. It already fails secure — the suggested fix is already there, one
  line above the quoted code.   

  3. "CSRF customRequestHeader !== '1' is bypassable."
  The check (ConsoleCsrfProtection.ts:37-41) is a 5-condition AND; '1' is just the outer layer. Real protection: a double-submit token compared with
  timingSafeEqual, bound to the server-side session hash, plus an Origin/Sec-Fetch-Site check. The '1' is the OWASP custom-header pattern (forces a CORS
  preflight) — not a secret, so bypassing it gains nothing.

  One point had merit: the rate-limit fail-open. The framing was off (a failing store fails closed — errors propagate to a 500), but it surfaced a real
  gap: admin step-up lacked a construction-time guard requiring a rate-limit store, unlike the DCR/local-password/magic-link paths. We've added that
  guard plus tests — so the review did drive a fix.

  Re-review with this in mind? Checking claims against the surrounding code (enclosing try/catch, multi-condition guards) before flagging would make the
  pass more reliable.

insomnolence:
## Remaining SonarCloud findings — reviewed, not actionable

The genuinely-actionable findings on this branch have been fixed. The ones still showing were each reviewed and are **false positives or unfixable-by-design**. We can't mark them Won't-Fix in SonarCloud from here, so this comment is the record of that review:

- **`S4325` "unnecessary type assertion" (×48)** — every one is a *required* cast, not a redundant one. The `as never` casts pass deliberately-minimal stubs where a full type is expected (test mocks like `{} as never`, and the `LowLevelMcpServer` → MCP SDK `Server` bridge in the server wiring); the `as unknown as Row[]` casts narrow Drizzle's loosely-typed raw `execute()` results. Removing any of them fails the TypeScript build.
- **`S1192` duplicated literals (×10)** — repeated values inside `CHECK` constraints in migration DDL. Raw PostgreSQL DDL has no constant mechanism to factor them out, and applied migrations are immutable.
- **`S2094` empty class (×3)** — intentional no-op stub classes implementing interfaces in tests.
- **`S6582` (×1, `InMemoryConsoleAccountAdminStore`)** — the suggested optional-chain rewrite would drop the non-null narrowing the next line relies on.
- **`S6551` (×1)** — the value reaching `String()` is never an object (guarded by the preceding `typeof` checks).
- **`S7924` (×1)** — a deliberate warning-amber accent color.

All are maintainability-only; the quality gate (reliability / security / hotspots / maintainability / duplication) is green.

PR #2275, opened 2026-06-15
fix: repair OAuth helper token storage and status reporting

### LINKED ISSUE #2262: Device-flow OAuth never completes: oauth-helper calls removed static TokenManager API; pending_token.txt fallback has no consumer
## Summary

GitHub device-flow OAuth (`setup_github_auth`) can never complete in any v2.x release. The OAuth helper successfully obtains the token from GitHub, but both storage paths are broken:

1. **Primary path throws on every call**: `oauth-helper.mjs:130` calls `TokenManager.storeGitHubToken(token)` as a **static** method. The v2.0.0-rc.1 platform rewrite (`62fb096c`) converted `TokenManager` to instance methods, but `oauth-helper.mjs` was never updated. The call throws `TypeError: TokenManager.storeGitHubToken is not a function` 100% of the time, silently swallowed by the bare `catch` at `oauth-helper.mjs:133`.
2. **Fallback path has no consumer**: the catch block writes the token to `~/.dollhouse/.auth/pending_token.txt` with a comment saying it's "for the MCP server to pick up" — but **no code anywhere in `src/` reads this file**. The token is written and orphaned forever. `getAuthStatus()` only checks the `GITHUB_TOKEN` env var and the encrypted token file, so the user remains "Not Connected" despite a valid token sitting on disk.

Net effect: the user authorizes on GitHub, the helper gets a valid `gho_` token, and auth still never completes. `check_github_auth` reports "In Progress" until the device code expires, then "Expired".

## Reproduction (against published `@dollhousemcp/mcp-server@2.0.32`)

```bash
cd ~/.npm/_npx/<hash>/node_modules/@dollhousemcp/mcp-server
node -e "
import('./dist/security/tokenManager.js').then(async ({TokenManager}) => {
  try { await TokenManager.storeGitHubToken('gho_fake_token_for_repro'); console.log('OK'); }
  catch (e) { console.log('FAILS WITH:', e.constructor.name + ':', e.message); }
})"
# => FAILS WITH: TypeError: TokenManager.storeGitHubToken is not a function
```

End-to-end: run `setup_github_auth`, authorize the device code on GitHub. Observe `~/.dollhouse/.auth/pending_token.txt` appears containing a valid `gho_` token (mode 0600), while `check_github_auth` still reports "Authentication In Progress" / "Not Connected".

## Secondary bugs: misleading diagnostics (found while chasing a phantom crash)

The diagnostics actively misdirect debugging — they made a working helper look like a crashed one:

- **Exit-after-success reported as crash**: the helper, by design, "polls GitHub for the OAuth token, stores it securely, and then exits" (`oauth-helper.mjs:6-8`). But `checkOAuthHelperHealth()` treats a non-alive PID during an unexpired device-code window as a crash: `oauth_helper_status` shows "⚠️ WARNING: Process appears to have stopped! ... may have crashed".
- **Contradictory status**: `oauth_helper_status` displays "**Status:** 🟢 ACTIVE - Authentication in progress" and "**Process Alive:** ❌ No (may have crashed)" in the same output (`GitHubAuthHandler.ts:365-380`).
- **Phantom log file**: helper logging is disabled unless `DOLLHOUSE_OAUTH_DEBUG=true` (`oauth-helper.mjs:55`), so `oauth-helper.log` never exists in a default install — yet troubleshooting output tells users to "Check the log file for errors" and its absence reads as crash evidence.
- **Stale state file**: `oauth-helper-state.json` is only cleaned up on the success path that can never be reached, so it lingers and keeps reporting a dead "in progress" session.

## Root cause timeline

- Pre-2.0: `TokenManager` had static methods; `oauth-helper.mjs` called them correctly.
- `62fb096c` (v2.0.0-rc.1, 2026-03-19): platform rewrite made `TokenManager` instance-based (DI'd `fileOperations`); `oauth-helper.mjs` was not updated.
- Broken in every v2.x release since, including current 2.0.32.

## Suggested fixes

1. In `oauth-helper.mjs`, instantiate the manager (`new TokenManager(...)` with whatever deps it needs) or expose a static/standalone `storeGitHubToken` entry point for out-of-process callers.
2. Either implement the `pending_token.txt` pickup in the server (e.g., in `getAuthStatus()`/`checkGitHubAuth()`: read, validate, store via TokenManager, delete file), or remove the dead fallback.
3. Fix diagnostics: distinguish "helper exited" from "helper crashed" (e.g., helper writes a completion marker / result file before exit); don't reference a log file that only exists in debug mode; clean up the state file on all terminal paths.
4. Add an integration test that exercises the helper's storage path against the built `dist/` so a static/instance API drift fails CI.

## Environment

- `@dollhousemcp/mcp-server@2.0.32` via `npx` (npm cache), macOS (darwin 25.5.0), Node from npx runtime
- Reproduced twice in a row in a live Claude Code MCP session (device codes E5FE-3996, B1BE-5639)

## Follow-up

Verify whether the HTTP integration branch is affected — its auth flow may bypass `oauth-helper.mjs` entirely or share the same TokenManager API drift.

jstar0:
## Summary

Fixes #2262.

This updates the detached device-flow helper so it stores OAuth tokens through the current `TokenManager` instance API, removes the unconsumed `pending_token.txt` success fallback, and records explicit terminal helper results for status reporting.

## Changes

- Instantiate `TokenManager` inside `oauth-helper.mjs` with the helper's file-operations adapter before storing device-flow tokens.
- Remove the plaintext `pending_token.txt` success fallback from the helper path and report storage failures as non-zero helper failures.
- Write a small non-sensitive `oauth-helper-result.json` on terminal helper paths and clean stale helper state when the helper finishes.
- Teach `GitHubAuthHandler` to report completed, failed, expired, denied, and timed-out helper results without inferring a crash from expected process exit.
- Avoid pointing users at a missing debug log when OAuth debug logging is disabled.
- Keep helper state/result/pid/log files aligned with `DOLLHOUSE_HOME_DIR` when it is configured.
- Add subprocess and handler regressions for token storage, terminal result files, state cleanup, and diagnostics.

## Verification

```bash
npm test -- tests/unit/auth/oauth-helper.test.ts --runInBand
npm test -- tests/unit/auth/oauth-helper.test.ts tests/unit/handlers/GitHubAuthHandler.test.ts tests/unit/auth/GitHubAuthManager.test.ts tests/unit/security/tokenManager.storage.test.ts --runInBand
npm run build
```

### DIFF
--- a/oauth-helper.mjs
+++ b/oauth-helper.mjs
@@ -27,6 +27,11 @@ const __dirname = dirname(__filename);
 const DEFAULT_POLL_INTERVAL = 5;
 const DEFAULT_EXPIRES_IN = 900; // 15 minutes
 const MAX_TOKEN_SIZE = 10000; // Maximum reasonable token size
+const DOLLHOUSE_HOME_DIR = process.env.DOLLHOUSE_HOME_DIR || homedir();
+const AUTH_DIR = join(DOLLHOUSE_HOME_DIR, '.dollhouse', '.auth');
+const PID_FILE = join(AUTH_DIR, 'oauth-helper.pid');
+const STATE_FILE = join(AUTH_DIR, 'oauth-helper-state.json');
+const RESULT_FILE = join(AUTH_DIR, 'oauth-helper-result.json');
 
 // Parse command line arguments
 const args = process.argv.slice(2);
@@ -51,7 +56,7 @@ if (!clientId || clientId === 'undefined') {
 }
 
 // Log file for debugging (optional, can be disabled in production)
-const LOG_FILE = join(homedir(), '.dollhouse', 'oauth-helper.log');
+const LOG_FILE = join(DOLLHOUSE_HOME_DIR, '.dollhouse', 'oauth-helper.log');
 const LOG_ENABLED = process.env.DOLLHOUSE_OAUTH_DEBUG === 'true';
 
 async function log(message) {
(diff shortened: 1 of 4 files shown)


mickdarling:
Thank you for the patch. I am going to create a more comprehensive version that includes a few other aspects. However, this was helpful. I appreciate the effort.  I'll reference this and you in the final PR.

mickdarling:
Thank you again for the report and patch. The root-cause analysis here was useful and helped confirm the broken helper/token-storage path.

Because this touches OAuth token handling, we are landing the fix through an internally maintained PR with broader diagnostics, stale plaintext fallback cleanup, and regression coverage for the helper boundary: #2323.

Closing this as superseded by #2323, not because the diagnosis was wrong. We reference this PR and credit your contribution in the replacement PR body.

mickdarling:
Superseded by #2323, which incorporates the root-cause fix with broader OAuth helper diagnostics and tests.

PR #2318, opened 2026-06-26
feat(web-console): user-owned integration vault, gated execution, and curated connect

insomnolence:
## Summary

Adds user-owned third-party integrations to the web console as a **data-driven** capability. A provider is defined by a declarative descriptor plus an OpenAPI spec — not bespoke per-provider code — so adding a standard OAuth2 or API-key service is configuration rather than a development cycle.

The model has three parts:

- **Credential vault.** Per-user OAuth tokens and API keys are encrypted server-side with per-user/per-provider AAD. They never reach the browser or the model.
- **One gated execution path.** A single `integration_request` tool resolves the caller's credential, injects it server-side, and forwards the call — constrained to the descriptor's host allowlist, with HTTPS-only, method gating, request/response size caps, outbound rate limits, an SSRF guard on the resolved host, a single 401 refresh-and-retry, credential-shaped response redaction, and an integration-aware audit trail. Write methods are escalated to the gatekeeper for approval; third-party responses are marked untrusted.
- **Agent-facing operation knowledge.** Named operations are derived from the stored OpenAPI spec, with scope-aware availability and `list_operations` / `describe_operation` discovery. A bounded, editable Skill element is generated as a convenience layer. Correctness lives in the tool, never in the generated prose — an edited or wrong skill can only ever produce a call the tool independently constrains.

On top of the REST path it adds promoted per-operation typed tools and an allowlisted remote-MCP bridge, both behind the same gateway/gatekeeper guarantees.

It also adds **curated, deployment-shipped providers**: descriptor seed files (data, loaded into the configured store at startup — works on both in-memory and database backends) plus deployment OAuth credentials supplied via environment, which activate the generic per-provider connect / callback / disconnect routes. Providers that ship without credentials configured stay dormant until a deployment supplies them.

GitHub's existing connect and portfolio-sync path is unchanged.

## Not in this change (planned follow-ups)

- User self-service authoring of their own providers (descriptor + spec management from the console).
- Operator runtime management of curated providers directly in the database.
- The Integrations UI surfacing the new providers (the tab still shows the existing GitHub card).
- Outbound SSRF guard hardening (canonical-IP classifier + resolve-once-and-pin across the data and token-endpoint paths) before untrusted user-authored descriptors are accepted.

## Testing

Full suite is green — unit, integration, e2e, and console-e2e. The assembled feature is exercised end to end through real dependency-injection wiring (not mocks), including multi-user credential isolation, response credential redaction, write-class approval, remote-MCP credential injection, and curated descriptor route registration on both auth strategies.


insomnolence:
Deferred follow-up work from this PR's reviews is tracked in:

- **#2321 — BYO authoring + spec management**: the remaining work to complete Integrations v2 (the self-service surface to author your own integration, vs. only connecting curated ones), plus the BYO-input acceptance gates (OpenAPI `$ref` depth-limit rejection, descriptor-list pagination, and folding the policy check inside the gateway so it can't be invoked un-gated).
- **#2320 — outbound SSRF resolve-once-and-pin + canonical-IP classifier**: closes the canonical-address classifier gap, the DNS-rebinding (TOCTOU) window, and the unguarded OAuth token-endpoint fetches. #2321 is blocked by this for the parts that accept untrusted hosts/URLs.

All of these are gated on accepting **untrusted bring-your-own descriptors/specs**, which this PR does not enable — it ships curated, deployment-controlled integrations only. The redirect-replay vector raised in review is already closed here (`redirect: 'error'` on every credential-bearing fetch).

Issue #2321, opened 2026-06-28
Integrations v2: BYO (bring-your-own) authoring + spec management


insomnolence:
## Summary
Add the self-service authoring surface that lets a user create and manage their **own** integration, completing Integrations v2. After PR #2318, users can only **connect** deployment-owned ("curated") integrations — they cannot author their own.

## Already shipped in PR #2318 (do not redo)
The Integrations v2 engine for curated integrations:
- Per-user encrypted credential vault (OAuth + static API key); credentials injected server-side only, never exposed to the browser or model.
- Gated outbound execution via the `integration_request` MCP tool (`IntegrationRequestGateway`): method/host allowlisting, HTTPS-only, request/response size caps, per-user rate limiting, response credential redaction, audit events, `redirect: 'error'` on credential-bearing fetches.
- OpenAPI-as-data operation discovery (`IntegrationOperationCatalog`: `ingest_openapi_spec`, `list_operations`, `describe_operation`) and a remote-MCP bridge.
- **Curated connect**: connect/disconnect deployment-owned integrations from the console.

The descriptor/spec stores and validation already exist and are reused here; what's missing is the user-facing authoring API on top of them.

## Scope
**Endpoints (console `/api/v1`, owner-scoped):**
- [ ] Create / update / delete a BYO descriptor (provider id, display fields, auth strategy, API hosts, OAuth public config, static-key injection, promotion config) — client secret encrypted at rest, never returned to the browser.
- [ ] Upload / replace / fetch the descriptor's OpenAPI spec.
- [ ] (optional) List derived operations / manage promotion config.

**Store work:**
- [ ] Add `delete(id, ownerUserId)` and owner-scoped `findById(id, userId)` to the in-memory and Postgres descriptor stores (today only `findVisibleByProvider`/`upsert` exist); add a Postgres migration if the delete path needs one.

**Wiring:**
- [ ] Populate the provider registry from authored descriptors so a newly authored BYO descriptor is connectable without a restart (per-request resolution or registry refresh on write).
- [ ] DTOs + privacy projectors exposing only display-safe fields — **never** `clientSecretCiphertext`, key versions, or any credential material.
- [ ] Register the route module + module-id gating; update the API contract + route manifest (manifest-parity test must pass).

**Acceptance gates — required before accepting untrusted BYO input:**
[...]

PR #2328, opened 2026-07-03
Integrations v2: BYO descriptor authoring + spec management

insomnolence:
Completes #2321 — the self-service BYO (bring-your-own) authoring surface that lets a user create and manage their own integration descriptors, finishing Integrations v2. After the curated-connect work, users could only connect deployment-owned integrations; this adds the authoring plane on top of the existing descriptor/spec stores and gateway.

## What landed

**Self-authorizing execution authorities.** The integration request gateway, remote-MCP bridge, and operation-catalog writes now fold the policy check inside the only invocation path via `Authorized*` facades: the raw authorities are removed from the module barrel and constructed only by the DI root, each facade requires its policy enforcer at construction, and a lint rule blocks raw construction elsewhere — an un-gated outbound authority is unrepresentable rather than merely discouraged. Session-start remote-MCP tool discovery is now gated behind a side-effect-free policy check that fails closed.

**BYO descriptor authoring API** (`/api/v1/me/integrations/descriptors`). Owner-scoped create/list/get/update/delete, with ownership forced to the caller. The OAuth client secret is write-only — accepted on create/update, encrypted immediately with context-bound AAD, and never serialized back (the DTO exposes only `has_client_secret`). A BYO descriptor cannot claim a provider id owned by a built-in or curated provider, and provider resolution prefers curated strictly over BYO so a user cannot shadow a deployment provider.

**OpenAPI spec management** (`/descriptors/{id}/spec`). Owner-scoped ingest/replace/fetch sharing the same fail-closed validation as the agent-facing ingestion (OpenAPI 3.x, local-`#/`-only `$ref`s rejected past the nesting-depth limit, host-allowlist, size cap). Responses carry spec metadata only; the stored spec stays server-side data.

**Per-request provider resolution.** Connect/status/callback/disconnect and token refresh resolve the descriptor from the store per request, so a runtime-authored BYO descriptor is connectable and refreshable without a server restart.

**Auth strategies.** `oauth2_authorization_code` and `static_api_key`, with HTTP Basic injection (`Authorization: Basic base64(username:password)`, encoded at injection time) for Twilio/Mailgun-class APIs.

**Hardening surfaced by review.** Reserved-id and registry-aware namespace protection so a descriptor cannot route a built-in provider's stored credential to a chosen host; deterministic curated-preferred resolution; correct client-secret rotation on partial updates; keyset descriptor pagination with the in-memory sort aligned to the cursor comparison.

## Notes

- No new runtime dependencies.
- Store deletion needs no migration — the descriptor→spec relationship already cascades via the existing foreign key.
- The route manifest-parity test is updated in lockstep with the new routes.


insomnolence:
Thanks for the review. Notes on the items that were addressed differently or deliberately left as-is, so they're not open questions at merge:

**Concurrent-create / transaction isolation.** `withSystemContext` runs its callback inside `db.transaction(...)`, so the descriptor `upsert`'s SELECT-then-INSERT/UPDATE executes transactionally, and the unique constraint on `(provider, ownership, owner_user_id)` guarantees no duplicate row. Under READ COMMITTED two simultaneous creates for the same provider can both pass the pre-check; the loser now trips the constraint and is translated to the same `409` the pre-check returns (rather than a 500) via the `isUniqueViolation` catch in `create()`. So the race is bounded to a clean conflict response, not a data or error-handling hazard.

**Special-use TLDs (`.test` / `.example` / `.invalid` / `.arpa`).** Deliberately not rejected at descriptor validation. Those TLDs are RFC-reserved specifically so tests can use them as stand-in hosts, and the suite relies on that (the wired harness and several fixtures use `.test`/`.example` as their API hosts). Blocking them at authoring would break those fixtures with no legitimate public host to migrate to, and the production benefit is marginal — those names never resolve in real DNS, so the outbound resolve-and-pin guard already fails them closed at request time. The private-network guards that matter (`.localhost`, `.local`, `.internal`, IP literals) remain enforced.

**`policyArguments` coercing non-object arguments to `{}`.** Not a policy/HMAC mismatch: remote-MCP tool arguments are an object by protocol, and the bridge's `readArguments` rejects a non-object (including a top-level array) with a `400` before any call executes. An array-shaped argument can never reach the tool, so there's nothing for the policy input to diverge from.

**Test / style suggestions** (a live two-connection concurrency test, a direct null-return test for the OAuth-without-ciphertext resolver path, and the stacked JSDoc block in `CuratedIntegrationProviders.ts`). All noted as reasonable, none blocking: the 409 translation is covered by a unit test on the mapping itself, and the null-resolver behavior is already exercised end-to-end by the "fails closed for a BYO OAuth descriptor without a stored client secret" case. These are opportunistic follow-ups rather than changes for this PR.

PR #2357, opened 2026-07-06
feat(web-console): collection catalog browse and install

### LINKED ISSUE #2317: Configure hosted portfolio tab to use collection.dollhousemcp source
## Context

During alpha testing of the hosted DollhouseMCP web console, the Portfolio tab does not appear to be configured to look at the expected collection source. The likely intended source is `collection.dollhousemcp`.

## Problem

The hosted server's Portfolio tab should surface the correct hosted/collection-backed portfolio data, but it currently appears disconnected from the expected collection configuration.

## Desired Behavior / Acceptance Criteria

- Confirm the intended collection source name/identifier for hosted alpha/beta deployments, likely `collection.dollhousemcp`.
- Configure the hosted web console Portfolio tab to read from the correct collection source.
- Ensure the configuration is explicit in hosted deployment env/config, not an implicit local/default assumption.
- Show a useful empty/error state if the collection source is unavailable or misconfigured.
- Add verification steps so deploy smoke tests can confirm the Portfolio tab is pointed at the intended collection source.
- Document the required hosted configuration for alpha, beta, and future production/enterprise deployments.

## Notes

This is distinct from graphical import/export and local sync. This issue is specifically about making sure the hosted Portfolio tab is wired to the correct collection-backed source.

insomnolence:
Closes #2317

Adds the community collection to the web console: browse/search the public catalog from the Portfolio tab and install elements into the signed-in user's portfolio, wiring the hosted Portfolio tab to the collection.dollhousemcp source.

## What's here

**Browse surface** — `GET /api/v1/collection/elements` (browse + search) and `GET /api/v1/collection/elements/:type/:name` (detail), served by a new `collection` BFF module over the existing collection engine (`CollectionIndexManager`, `CollectionSearch`, `PersonaDetails`). The catalog is public data, but the routes stay session-gated because every request can drive server-funded outbound GitHub traffic. A new `public_catalog` privacy class and a dedicated `collection_fetch` rate policy (per-session and per-user, store-backed) cover the surface, and the module registry refuses `public_catalog` routes outside `/api/v1/collection`.

**Install** — `POST /api/v1/me/portfolio/from-collection` (self-private, idempotency required). It lives under the portfolio path rather than `/collection` because installing creates a private copy of a catalog element. Fetch + validation happen through a new no-write seam on the shared installer (`ElementInstaller.fetchAndValidate`), and the write goes through the same manager-backed store the portfolio CRUD routes use — DB-mode deployments persist to Postgres with per-user scoping and no filesystem fallback. Installs apply the same pre-write payload validation as direct portfolio creates, so an out-of-contract catalog element is rejected with a 422 before anything persists. All six element types are supported, including pure-YAML memories with their own secure validation branch.

**UI** — the Portfolio tab's Collection source is now live: lazy-loaded catalog, COLLECTION badges, always-visible Install buttons (hidden on browse-only deployments via a new `install_enabled` list field), a detail modal for both sources, and distinct not-enabled / degraded / error states with retry.

**Security hardening** — SSRF guards reject raw and percent-encoded traversal in collection paths before URL normalization; the browse cache is store-backed when a cache store is configured (filesystem only as the no-store fallback); collection section/type inputs validate against allowlists.

## Deployment

Everything is behind `DOLLHOUSE_WEB_CONSOLE_COLLECTION_ENABLED` (default **off**) — browse spends the deployment's shared GitHub API budget, so operators opt in deliberately. The install route additionally requires the portfolio-write flag. The replacement-readiness check treats `collection` as an optional module, so default deployments boot with the flag off. Hosted deployments should set a **zero-scope** `GITHUB_TOKEN` (public catalog reads need no scopes; `public_repo` would grant public-repo write) to raise the GitHub budget from 60 to 5,000 requests/hour — startup logs a warning on non-loopback binds without one.

## Testing

Unit coverage for the module, install service, installer seam (all six types, extension enforcement, traversal and security rejections), rate limiter, registrar gating, and readiness matrix; full unit/integration/e2e suites green. Browse → install → persisted-to-Postgres → visible-in-portfolio verified end-to-end in a browser against a production-mode boot.

### DIFF
--- a/docs/guides/v2-migration-guide.md
+++ b/docs/guides/v2-migration-guide.md
@@ -322,6 +322,7 @@ New configuration options in v2.0:
 | `DOLLHOUSE_CONSOLE_TOKEN_FILE` | `~/.dollhouse/run/console-token.auth.json` | Optional override for the console token file path. |
 | `DOLLHOUSE_CONSOLE_ROTATION_REQUIRE_CONFIRMATION` | `true` | Require out-of-band confirmation (Phase 2: OS dialog or TOTP) for token rotation. Set `false` for headless CI. |
 | `DOLLHOUSE_INTEGRATION_DESCRIPTOR_SEED_DIR` | *(unset)* | Directory of curated integration descriptor seed files (data, not code) loaded into the descriptor store at web-console bootstrap. Per-provider deployment OAuth credentials are read separately from `DOLLHOUSE_INTEGRATION_<ID>_CLIENT_ID` / `_CLIENT_SECRET`. |
+| `DOLLHOUSE_WEB_CONSOLE_COLLECTION_ENABLED` | `false` | Enable collection catalog browse routes (`/api/v1/collection/*`) and the Portfolio tab's Collection source. Drives server-funded outbound GitHub traffic under a shared rate budget — opt in deliberately, and set a zero-scope `GITHUB_TOKEN` to raise the budget from 60 to 5,000 req/hr. Install into the portfolio additionally requires `DOLLHOUSE_WEB_CONSOLE_PORTFOLIO_WRITE_ROUTES_ENABLED`. |
 
 ---
 
--- a/src/cache/CollectionCache.ts
+++ b/src/cache/CollectionCache.ts
@@ -5,7 +5,8 @@
 import * as path from 'path';
 import { logger } from '../utils/logger.js';
 import { SecurityMonitor } from '../security/securityMonitor.js';
-import { IFileOperationsService } from '../services/FileOperationsService.js';
+import type { IFileOperationsService } from '../services/FileOperationsService.js';
+import type { ISharedCacheStore } from '../storage/sharedCache/ISharedCacheStore.js';
 
 export interface CollectionItem {
   name: string;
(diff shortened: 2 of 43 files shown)


insomnolence:
Follow-up on the four low-severity observations from the latest review, so their disposition is on record before merge:

**Percent-encoded hostnames (`GitHubClient.assertUrlAllowed`)** — Covered. The test suite now includes a case rejecting a percent-encoded, non-allowlisted hostname. The suggested bypass isn't reachable: the URL parser canonicalizes the hostname the same way at allowlist-check time and at connection time, so the host that gets checked is always the host that gets fetched.

**Rate-limit store wiring** — Verified fail-closed. When collection routes are enabled without a `RateLimitStore` registration, the registrar refuses to start (an existing test pins that error). Startup also warns when the configured backend is process-local, so multi-replica deployments get pointed at a shared backend.

**`loadCollection` try/catch structure** — No change; the review itself notes the current form renders the error state correctly and flags this as style only.

**`'skills'` fallback in the list projector** — Intentional forward-compatibility: an unknown catalog type shouldn't break the UI. It's unreachable in practice because the service layer filters to the supported element types before serialization.

PR #2358, opened 2026-07-07
feat(web-console): API contract conformance — error envelope, profile fields, pagination

insomnolence:
Locks down the three response-shape contracts a production UI (and any future API client) builds against, while the `/api/v1` surface still has a single first-party consumer and shape changes are free.

## Error envelope

Errors returned from inside route handlers previously shipped as plain `application/json` with `type: "about:blank"` and no correlation id, while middleware errors (authentication, CSRF, authorization) correctly used RFC 9457 `problem+json`. The route-execution kernel now lifts handler error bodies through the same problem builder — typed `type` URI derived from the error `code`, `instance` populated with the request correlation id, extension members such as validation `issues` preserved. One enforcement point covers every module, present and future; the module-level error helpers are unchanged. The browser-navigation HTML error page keeps precedence for document requests.

## `GET /auth/me` profile fields

The response now carries `display_name`, `email`, and `auth_methods`, served from the same account store `/me/profile` reads — no parallel profile path. Null-safe when the principal row is gone, so a session that outlives its principal can still resolve and sign out. The field is the `auth_methods` array (the account's linked methods, matching `/me/profile`) rather than a singular `auth_method`: the session record carries no per-session method attribution, so a singular value would be fabricated.

## Pagination: two list families

List endpoints previously spanned five envelope shapes, including two broken ones — `/me/logs` reported `has_more` with no token to fetch more, and the system-metrics endpoint exposed its pagination as a raw integer `offset`. The surface now uses exactly two families:

- **Cursor** — `{items, page: {limit, cursor, next_cursor}}` for unbounded/append-only surfaces. Cursors are opaque and filter-bound; `next_cursor: null` is the exhaustion signal; there is no `total` (counting an unbounded set is either unaffordable or meaningless). `/me/logs` and `/admin/operate/metrics/system` converge into this family via a shared cursor codec.
- **Snapshot** — `{<noun>: [...]}` for bounded sets, never a bare array, with the envelope key reserved for future evolution. The three runtime-session lists (which returned bare, unevolvable JSON arrays) converge here; the self-session list also gains its previously missing privacy-projector declaration.

The collection catalog's page-number form is grandfathered as a documented exception (a cached remote snapshot where `total` is free and the interaction model is browse-then-filter). The never-implemented page-based + `Link`-header model is withdrawn from the contract. Consuming UI modules are updated in the same change, so the bundled console is never off-contract.

## Testing

Unit coverage for the kernel problem lift, the cursor codec, cursor page-walks on both converged endpoints, and the envelope shapes; new end-to-end assertions pin the problem envelope (including `instance` = `X-Correlation-Id`) and both list families over real HTTP against a production-mode boot. Full unit/integration/e2e suites green; Logs and Sessions tabs verified live in a browser.

### DIFF
--- a/src/web-console/WebConsoleRegistrar.ts
+++ b/src/web-console/WebConsoleRegistrar.ts
@@ -500,6 +500,7 @@ export class WebConsoleRegistrar {
         loginTransactions: stores.loginTransactionStore,
         sessionStore: stores.sessionStore,
         identityResolver: stores.identityResolver,
+        accountAdminStore: stores.accountAdminStore,
         opaqueValues,
         secretEncryption,
         publicBaseUrl: integrationPublicBaseUrl,
--- a/src/web-console/auth/ConsoleBffAuthModule.ts
+++ b/src/web-console/auth/ConsoleBffAuthModule.ts
@@ -502,6 +502,9 @@ function invalidCapability(): ConsoleHandlerResult {
   return {
     status: 400,
     body: {
+      type: 'about:blank',
+      title: 'Invalid request',
+      status: 400,
       code: 'invalid_capability',
       detail: 'Step-up requires a valid administrative console capability.',
     },
--- a/src/web-console/auth/ConsoleBffAuthModule.ts
+++ b/src/web-console/auth/ConsoleBffAuthModule.ts
@@ -11,6 +11,7 @@ import {
 } from '../middleware/ConsoleCookies.js';
 import type { IConsoleOpaqueValueService } from '../security/ConsoleOpaqueValues.js';
 import type { ISecretEncryptionService } from '../security/SecretEncryption.js';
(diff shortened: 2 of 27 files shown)


insomnolence:
Disposition of the one remaining observation from the latest review:

**E2E assertion for `instance` = `X-Correlation-Id`** — Already covered. The portfolio e2e spec asserts exactly this over real HTTP on a 404 (`me-portfolio.e2e-spec.ts`: `expect(problem.instance).toBe(res.headers.get('x-correlation-id'))`), shipped together with the error-envelope change itself, and two unit tests pin the same mapping on the step-up error paths. The review checked a different spec file and missed it.

With that, every observation across all three review rounds is fixed, tested, or dispositioned.

PR #2364, opened 2026-07-13
feat(web-console): paginate admin lists and add UI discovery endpoints

insomnolence:
## Summary

The cross-tenant users directory and cross-user active-sessions list now use cursor pagination with server-side search/filter instead of a capped one-shot response, so they stay usable past the old row cap. The console also serves the metadata a UI needs to build against without hardcoding server truth.

**Lists → cursor pagination + filter**
- Users directory: search by name/email, filter by role and enabled/disabled.
- Operational sessions: filter by user and status.

These two lists grow with the tenant/session population rather than one person's activity, so a bounded snapshot no longer fits. Per-user lists (a user's own sessions, their portfolio) are unchanged.

**New discovery / utility endpoints**
- `GET /api/v1/me/manifest` — the registered routes and their required capabilities, so the SPA can feature-detect instead of probing for 404s.
- `GET /api/v1/me/role-catalog` — the admin roles and the capabilities each grants, so a permissions UI stops hardcoding the mapping.
- `GET .../sessions/commands/:command_id` (self + operate) — poll an async session-termination to completion.
- `GET .../descriptors/:id/spec/operations` — enumerate the operations a BYO integration spec exposes.
- `GET /api/v1/admin/accounts/identities/unlinked` — logins not yet attached to a user (the identity-link picker source).

**Health honesty**
- The operations database health check now runs a real query, so a configured-but-unreachable database reports degraded instead of green.

Migration `0044` adds the indexes backing the paginated queries.

## Testing
Full unit, integration, and security suites pass; lint and typecheck are clean. New behavioral tests cover the endpoints, keyset pagination (tie boundaries and multi-page traversal), the filters, the cross-user ownership guard on command status, and the database-degraded health path.

### DIFF
--- a/src/database/migrations/0044_web_console_list_keyset_indexes.sql
+++ b/src/database/migrations/0044_web_console_list_keyset_indexes.sql
@@ -1,0 +1,20 @@
+-- Keyset-pagination indexes for the Family-B console list surfaces that scale with
+-- the tenant/session population (see API contract §5.3). Each index matches the exact
+-- ORDER BY + row-value cursor predicate of its query so pagination stays O(limit),
+-- independent of table size, instead of the prior full-scan + sort under a hard cap.
+
+-- Users directory: `ORDER BY created_at ASC, id ASC` over live accounts, cursor `(created_at, id) > ?`.
+CREATE INDEX IF NOT EXISTS "idx_users_created_at_id_active"
+  ON "users" ("created_at", "id")
+  WHERE "deleted_at" IS NULL;
+
+-- Cross-user operational sessions: `WHERE status=? ORDER BY last_active_at DESC, session_id DESC`,
+-- cursor `(last_active_at, session_id) < ?`. `lease_until > now()` stays a residual filter.
+CREATE INDEX IF NOT EXISTS "idx_runtime_session_presence_active_ordering"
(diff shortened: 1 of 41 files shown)


insomnolence:
Following up on Claude's review comments — dispositions on the remaining (all Low/Informational) items:

- **`search`/`sub` length ("could 500"):** these already return a clean **400**, not a 500. `ConsoleStoreValidationError` (e.g. `search` > 128 in `validatePrincipalDirectoryQuery`) maps to `400 invalid_request` in `problemForConsoleError`, which the secured router applies to handler errors. An overlong `sub` isn't a crash path either — it simply doesn't match the `varchar(320)` column and returns an empty page. A parse-layer cap would change only the message, not the status, so leaving it.
- **`auditOperations` empty-string (`AccountAdminModule.ts`):** inert here — all 22 routes in the module declare an `auditOperation`, so `?? ''` never fires and no empty-id entry is produced.
- **`revokeAllAccountSessions` sequential loop:** intentional. Bulk revoke is bounded at 500 and admin-initiated; sequential creation keeps one write in flight rather than fanning 500 concurrent inserts at the connection pool.
- **`pg_trgm` index / parser-idiom consistency:** reasonable follow-ups, out of scope here.

The `user_id` parse-layer guard is in `af3a2573`.

PR #2365, opened 2026-07-14
chore: merge develop into codex/hosted-http-integration

insomnolence:
Syncs `develop` into `codex/hosted-http-integration`.

This lands as a **real two-parent merge commit** (parent 1 = the integration
branch tip, parent 2 = the `develop` tip), so the sync is recorded in history
**as a merge** rather than a flattened/squashed commit. Please merge with a
method that preserves that merge commit (merge commit, not squash) so the
develop ancestry is retained on the integration branch.

Develop's fixes were **manually ported into the integration architecture**
rather than taking their pre-refactor implementations:

- **Memory persistence** — the serialized-memory size contract, persistability
  preflight, failure-ledger recovery/rollback, and search-index cleanup were
  ported into the extracted `MemorySaveHandler` and storage-aware
  `MemoryManager`, keeping session-scoped, per-user isolation.
- **OAuth / token handling** — the detached device-flow helper now hands the
  token to the session process through an encrypted, flow-bound handoff; the
  server stores it via the per-user `ITokenStore` (file or database) and
  verifies retrieval. Flow-id correlation, message redaction, and stale-artifact
  cleanup are included.
- **Setup / path / publishing** — NVM launcher self-heal (retaining the
  `PackageResourceLocator` path model), canonical path containment, and the
  publish-channel guards and workflows.

Release version advances to `2.0.38`.

Integration architecture is preserved throughout (extracted handlers,
dependency injection, per-user isolation, `ITokenStore`); generated artifacts
are regenerated from the resolved manifests rather than hand-merged.


insomnolence:
**Review findings — disposition**

**Not applicable**
- **`verify-publish-channel.mjs` beta regex** — `/-beta(\.|$)/` does not match `2.0.0-betaX`; the group requires `.` or end-of-string after `beta`, so `betaX` falls through to the unrecognized-prerelease branch and is refused (fail-closed). No change needed. This file is also unchanged from `develop` in this sync.
- **Flow-id mismatch test for `importCompletedOAuthHandoff`** — already covered in `GitHubAuthHandler.test.ts` ("does NOT import when the result flowId does not match the state").

**Pre-existing / out of scope for this sync**
- **`(spawnError as any)` / `(error as any)` casts** — pre-existing on the integration branch (prior high-complexity refactor), not introduced by this merge.
- **`device_code` passed as a CLI argument** — the existing device-flow helper contract on the integration branch; unchanged here.

**Deferred (hardening, tracked separately)**
- **Startup warning when `DOLLHOUSE_TOKEN_SECRET` is unset** — the at-rest handoff is guarded by 0600/0700 filesystem permissions; the warning is operator hygiene, not a correctness gap.
- **`flushPendingSaves` shutdown-context test** — the shutdown path's best-effort behavior is documented in code; the durable fix (session/user context propagation into shutdown) belongs with the broader multi-user hardening.

SonarCloud: 0 new issues.

PR #2376, opened 2026-07-28
feat(web-console): add custom integration authoring

insomnolence:
## Summary

The console's Connections list was a hardcoded GitHub-only array with a fixed
action-route table. This replaces it with a catalog built from the configured
provider statuses combined with the visible descriptor pages, and adds a
separate surface for authoring your own integrations.

Users can now define an integration, import its OpenAPI definition, and choose
which of the discovered operations are exposed as tools — without leaving the
console or hand-writing configuration.

## Connections catalog

Built from `GET /me/integrations` plus visible descriptors, so a runtime-authored
provider that isn't in the boot-time registry still appears. Supports literal and
parameterized `/:provider` routes, generic OAuth redirects, static API-key and
Basic credentials, and confirmed disconnects. Existing GitHub behaviour is
unchanged.

## Custom integration authoring

Create, edit, and delete owner-scoped descriptors with guided fields for
identity, allowed hosts, and authentication. Curated descriptors stay read-only.
OpenAPI definitions can be imported by file or paste, and discovered operations
are listed with method, path, and read/write classification.

Everything is expressed as ordinary form controls — operation promotion, OAuth
token-exchange settings, and the account-label mapping included. Operation
selection lives on the API-definition screen rather than the create form, because
the operation IDs only exist once a spec has been imported; asking for them
earlier is not answerable.

Selecting an operation turns it into its own named tool. Unselected operations
remain reachable through the general-purpose request tool. Because the stored
promotion settings are replaced wholesale by the API and are now edited from two
screens, each screen carries the other's portion through unchanged so that saving
in one place cannot silently discard the other.

## Credential handling

Client secrets and passwords are never rendered back — the DTO only exposes
`has_client_secret`. Plaintext is never held in module state, password fields are
cleared after submission and on teardown, and the submitted credential object is
wiped on both the success and failure paths. Leaving a secret blank while editing
preserves the stored value rather than clearing it.

## Testing

Console API E2E covers descriptor CRUD, DTO redaction, spec ingestion, derived
operations, and cross-user denial. Browser E2E covers the connect journey,
authoring with a write-only client secret, OpenAPI import, and operation
selection. The wired integration suite exercises BYO authoring through the
gateway.

### DIFF
--- a/src/web-console/ui/integration-descriptors.js
+++ b/src/web-console/ui/integration-descriptors.js
@@ -19,10 +19,12 @@ export function createIntegrationDescriptorManager(options) {
     descriptors: [],
     loading: false,
     error: false,
+    truncated: false,
     mode: 'list',
     editing: null,
     spec: null,
     operations: [],
+    specVersion: 0,
   };
 
   manager.host.addEventListener('click', event => onClick(manager, event));
@@ -30,14 +32,16 @@ export function createIntegrationDescriptorManager(options) {
   manager.host.addEventListener('change', event => onChange(manager, event));
 
   return {
-    setDescriptors(descriptors, error = false) {
+    setDescriptors(descriptors, error = false, truncated = false) {
       manager.descriptors = descriptors;
       manager.error = error;
+      manager.truncated = truncated;
       if (manager.mode === 'list') render(manager);
     },
     show() {
       manager.mode = 'list';
       manager.editing = null;
+      manager.specVersion += 1;
       render(manager);
     },
   };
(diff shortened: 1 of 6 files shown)


insomnolence:
We won’t make further changes for the remaining findings:

- `formatBytes(-1)` returns `0 B`, not `-1 B`. The `Math.max` guard applies whenever `bytes < 1024`, which includes negative values, so the suggested test would assert behavior the function does not have.
- `asObject` is defined in a single module; there is no duplicate to extract.
- The `problemDetail` duplication is three identical lines, and extracting them changes no behavior.
- The mock not counting a document without `openapi` or `swagger` as a stored spec is intentional. The service rejects those documents too.
- The in-flight spec request on delete and the abort-versus-error note are informational. The version guard already prevents stale responses from being applied.

All checks and relevant tests pass, so we consider the PR ready.

PR #2378, opened 2026-07-28
feat(web-console): add audit, security, and account governance

insomnolence:
## Summary

Adds audit, security administration, and account governance surfaces to the web console, giving elevated administrators the remaining read and control workflows the backend already exposed.

## What changed

- Adds a capability- and route-gated Audit tab covering admin action, approval, and authentication logs, with cursor paging and record detail.
- Reports audit integrity as recorded, including failed verification and approval records that cannot be verified.
- Adds a bounded audit export that ends on the server's terminal frame, caps accumulated records, and reports an incomplete file rather than presenting a partial export as whole.
- Omits filter controls from the audit surface, which the list routes do not implement.
- Adds a capability- and route-gated Security tab for signing-key lifecycle and the authentication policy.
- Adds key rotation, retirement, and deletion with consequence-stating confirmations and synchronous receipts.
- Defers the hard-delete grace window to the server, offering the force override only once the server reports that conflict.
- Presents the enforced authentication protections as stated invariants and limits editing to the maximum elevation length under its ETag.
- Adds allowlist, identity-triage, correlation-lookup, and bootstrap sections beside the existing account list, which is unchanged.
- Limits allowlist editing to the entry note, matching what the API accepts.
- Reports refusals requiring fresher elevation in place, since audit detail, the export, and every security mutation sit on stricter tiers than the surfaces that reach them.

## Test coverage

Beyond the expected success paths, browser tests assert the behavior that is easy to get wrong:

- Declining the hard-delete grace override leaves the key in place and sends no forced request.
- A rejected policy write leaves the stored policy unchanged rather than partially applied.
- An elevation refusal stays on screen after transient notifications clear.
- A partially entered allowlist value survives a background reload.
- Editing an allowlist entry sends only its note, never the immutable kind or value.
- The audit panel exposes no input controls, so filters the backend cannot serve are not added later.
- The existing account list renders unchanged after navigating the new sections.

## Verification

- ESLint and local Sonar lint passed for the complete touched-file set.
- Route-manifest parity test passed: 1/1.
- Authenticated browser suite passed: 21/21 across audit, security, accounts, users, and self-service.
- Allowlist create and delete additionally exercised against a live console instance.

### DIFF
--- a/src/web-console/ui/accounts-admin.css
+++ b/src/web-console/ui/accounts-admin.css
@@ -182,3 +182,14 @@
 }
 
 .acct-notice--error { border-color: rgba(239, 68, 68, 0.35); }
+
+.acct-note-input {
+  width: 100%;
+  padding: 4px 7px;
+  border: 1px solid var(--line);
+  border-radius: var(--radius-sm);
+  background: var(--paper);
+  color: var(--ink-900);
+  font: inherit;
+  font-size: var(--step--1);
+}
--- a/src/web-console/ui/accounts-admin.css
+++ b/src/web-console/ui/accounts-admin.css
@@ -1,0 +1,184 @@
+/* Account governance surfaces inside the Users tab: allowlist, identity triage,
+   and bootstrap status. The `ua-nav` rules belong here rather than to the
+   account list, which keeps its own styles in console.css. */
+
+.ua-nav {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 6px;
+  margin: 0 0 14px;
+  border-bottom: 1px solid var(--line);
+}
+
+.ua-nav-item {
+  padding: 7px 13px;
+  border: 0;
+  border-bottom: 2px solid transparent;
+  background: none;
+  color: var(--ink-500);
+  font: inherit;
+  font-size: var(--step--1);
+  cursor: pointer;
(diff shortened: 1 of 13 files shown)


insomnolence:
We won’t make further changes for the remaining findings:

- Export cancellation is covered. `console-auth.pw-spec.ts:610` cancels an in-progress export and asserts no download is offered; it drives the `exportTerminates` flag the review identified in the harness.
- The list-level elevation refusal is covered. `console-auth.pw-spec.ts:627` sets `listStatus` to 401 and asserts the panel reports it rather than remaining on its loading state.
- The `exportStream.stop()` sequence is the reverse of the description: `stop()` is called first and `finishExport` clears the handle afterwards. The comment at that call site records why stopping from inside the callback cannot re-enter it.
- `isGraceConflict` cannot use a structured code today. All three delete refusals return `conflict`, and only the grace case can be forced, so the message is currently the only way to tell them apart. Worth a follow-up to add a distinct code server-side, at which point the client check should move to it.
- The `users-admin.js` notes — the 200-user cap, reloading on tab activation, and the absence of a `destroy()` export — predate this change and are unchanged by it.

All checks and relevant tests pass, so we consider the PR ready.

Issue #2427, opened 2026-07-31
Stale agent activation after execute_agent never completes locks entire session (field incident 2026-07-30)


mickdarling:
## Incident

2026-07-30, Codex session `6252f7fb-e39d-4e21-a56a-eb63aa30e5fc` on the macOS Dev Work VM.

**Timeline (UTC, from `~/.dollhouse/logs/application-2026-07-30.log` and `security-2026-07-30.log`):**

- 20:26:32 — `execute_agent` runs `code-reviewer` (auto-confirmed, MODERATE risk=70). Agent activates. Its `tools.allowed: [read_file, glob, grep, list_directory]` is synthesized into a session-wide deny list by AgentToolPolicyTranslator (#449).
- 20:26:34 — two seconds later, `evaluate_permission` denials begin (WebUI/Gateway polling). No `complete_execution` / deactivation event ever appears in the log.
- 20:26 → 22:06 — ~100 minutes of denials on everything outside the agent's allowlist: `addEntry` (memory appends silently blocked), `get_active_elements`, `evaluate_permission`, and `deactivate_element` itself.
- 22:06:37 — the client model correctly invoked `release_deadlock` (exempt from deny synthesis, working as designed). Verification dialog displayed; session paused overnight awaiting the human code.

## Root cause

The agent execution ended (or died) without the agent element ever being deactivated. A stale active agent with an allowlist-inversion sandbox policy locks the whole session indefinitely.

## Proposed fixes (any/all)

- Auto-deactivate the agent element when execution completes, aborts, times out, or the executing context disappears.
- TTL on synthesized execution-scoped policies — a sandbox that outlives its execution should expire, not persist.
- Scope the synthesized policy to the agent's own execution context rather than the whole session's Gatekeeper decisions.

## Related

- Escape hatch worked as designed (#2026 flow); the failure is that it was needed at all.
- Companion issue: synthesized deny list blocks its own documented remedy (filed separately).

PR #2441, opened 2026-08-04
hotfix: recover stale agent execution policies safely

### LINKED ISSUE #2428: AgentToolPolicyTranslator synthesized deny list blocks its own documented remedy (deactivate_element, get_active_elements)
## Problem

When an agent's `tools.allowed` config is inverted into a deny list (#449), the synthesized list includes `deactivate_element` and `get_active_elements`.

The Gatekeeper deny error then says:

> Suggestion: Deactivate the element "code-reviewer" to proceed with this operation

…but `deactivate_element` is itself denied by the same policy. The suggested remedy is circular. `get_active_elements` being denied also blinds the session — you cannot even see *what* is locking you out.

Observed live in the 2026-07-30 incident (#2427): the only way out was full `release_deadlock` (deactivate everything + clear persisted state), which is disproportionate to "one stale agent needs deactivating."

## Proposed fix

Extend `EXEMPT_OPERATIONS` in `AgentToolPolicyTranslator.ts`, with different treatments:

- `get_active_elements` — exempt outright. Read-only introspection; being unable to see the lockout source is pure harm.
- `deactivate_element` — do **not** flat-exempt (a sandboxed agent could free itself). Instead route it through a **confirm/verification gate**: deactivating the sandboxing element requires the human verification-code flow, same pattern as `release_deadlock` but scoped to one element instead of nuking all activation state.

This gives a proportionate escape: human-approved single-element deactivation, with the full deadlock relief remaining as the last resort.

mickdarling:
## Summary

- preserve READ restrictions on `get_active_elements` while an agent sandbox is active
- keep `abort_execution` ungatable so explicit and synthesized policies cannot disable the narrow recovery path
- recover a stale in-memory execution policy only after durable state confirms no active goal remains or the agent itself is definitively absent
- revalidate durable goal state immediately before deleting a stale policy, including when a concurrent restart persists but policy tracking fails
- track bounded execution generations per agent in the shared `AgentManager` entry point, covering MCP-AQL and legacy `execute_agent` callers without cross-agent interference
- canonicalize in-memory execution-policy identities so equivalent agent-name aliases share one lifecycle entry
- confine strict durable-state reads to stale-policy recovery so normal `getState` behavior remains compatible
- hydrate flexible matches from the caller-requested state identity, never the legacy candidate filename
- bypass cached state during strict recovery checks and fail closed when a flexible match has no requested state file
- distinguish an absent state file from an unavailable state directory before accepting empty durable state
- propagate strict requested-state and flexible agent lookup failures so storage errors cannot masquerade as deletion
- request strict portfolio listing for recovery so `ENOENT`, `EACCES`, and `EPERM` cannot become an empty candidate set
- never recreate a missing agents directory during the security-sensitive recovery lookup
- return a successfully matched legacy-named agent despite unrelated candidate failures, but fail closed when no safe match can be established
- keep definition-only fallback candidates out of the shared cache and cache direct definitions only after hydration succeeds
- preserve a newer policy if the same agent restarts while stale-state lookup is in flight
- allow blocked agents to abort safely while preserving the independent DangerZone verification block
- include the scoped `abort_execution` remedy in Gatekeeper denial errors
- retain the denial on `deactivate_element`, so a sandboxed agent cannot silently free itself
- emit a distinct `AGENT_POLICY_RECOVERED` audit event and document the optional `recoveredStalePolicy` result field

## Root cause

An agent execution could disappear after its durable goal ended, or after the agent was deleted, while its in-memory policy remained active. The prior denial message suggested deactivation even though the sandbox could deny that same operation.

This hotfix provides a narrow recovery path without weakening unrelated policies, adding a TTL, or clearing every active element. Only two successful no-goals lookups, including one immediately before deletion, or definitive agent absence permit stale-policy removal. A bounded per-agent generation observation catches a concurrent restart before its durable goal or replacement policy exists, while avoiding false conflicts from unrelated agent executions. Execution policy keys use the same canonical identity as durable state and generation tracking, so an equivalent alias cannot strand a policy entry. Flexible matches preserve candidate-file state for ordinary reads and load durable state from the requested logical identity during strict recovery, and strict state-file or fallback-listing failures propagate instead of being synthesized as `ElementNotFoundError`; the security-sensitive lookup neither creates missing storage nor converts filesystem errors into an empty portfolio. Candidate loads are collected so a valid match remains usable despite an unrelated malformed agent, while any unresolved load failure prevents a false not-found result. Map-entry identity protects normally tracked restarts, and abort never clears a DangerZone verification requirement.

## Validation

- focused policy and policy-sanitization unit suites: 27 tests passed
- AgentManager load/read unit suites: 75 tests passed
- `agent-policy-enforcement.test.ts`: 26 integration tests passed, including READ restriction, explicit-abort denial, deleted-agent recovery, same-agent restart protection, unrelated-agent isolation, alias and sanitized identity isolation, DangerZone preservation, stale-cache durable completion, save-conflict fail-closed behavior, active-abort restart protection, and lookup-failure coverage
- focused ESLint passed
- `tsc --noEmit` passed

Fixes #2428
Mitigates #2427; automatic cleanup when an execution context disappears remains tracked there.


mickdarling:
@codex review

mickdarling:
@codex review

mickdarling:
@codex review current head bfc7931c

mickdarling:
@codex review current head 8916a763

mickdarling:
@codex review

mickdarling:
@codex review

PR #2442, opened 2026-08-04
hotfix: preserve append-only memory writes after scanner changes

### LINKED ISSUE #2440: Append-only memory becomes permanently unwritable when existing entries trip the content scanner (addEntry rejects with 'Malicious YAML content detected')
## Symptom

Every addEntry to the memory 'accelerator-directory' (16 entries, ~52KB, created and appended normally through MCP operations over the past week) now fails with:

    Entry NOT saved to memory 'accelerator-directory': Malicious YAML content detected

The failure is independent of the new entry's content. A one-line plain-ASCII test entry ("HF0 test entry A: F26 September 2026 batch cutoff 2026-08-15 UTC, verified first-party 2026-08-01.") is rejected on this memory, and the identical entry saves successfully to a freshly created memory. So the scanner is rejecting the memory's existing stored content during whole-file revalidation on save, not the incoming entry.

## Why this is bad

- Memories are append-only by design (no edit_element), so once historical content trips the scanner there is no MCP path to fix the memory: you cannot append, you cannot edit, and the guidance says never to touch element files directly. The memory is effectively bricked for writes while remaining readable.
- The entries now being flagged were all accepted by the same pipeline when they were written. Either the scanner tightened, or serialized escaping of older entries re-reads as something the scanner dislikes (possibly related to #2175's sanitizer behavior on technical content).
- The error gives no indication of which entry or pattern tripped detection, so users cannot even identify the offending content to work around it.

## Repro (Mick's portfolio)

1. Memory 'accelerator-directory' in the local portfolio (appended successfully as recently as 2026-07-29).
2. mcp_aql_create addEntry with any content, e.g. { element_name: "accelerator-directory", content: "plain ascii test" }.
3. Fails with Malicious YAML content detected. Same call against a new memory succeeds.

## Suggested direction

- Validate the incoming entry, not the whole existing file, on append; or if whole-file validation must run, treat already-persisted entries as trusted (they passed validation at write time) and only hard-fail on the delta.
- Include the failing entry id / byte range / matched rule in the error so a flagged memory is diagnosable.
- Provide a supported remediation path for a memory whose historical content is flagged (e.g. a quarantine-and-continue mode), since append-only + no-file-edits currently means permanent write lockout.

Workaround in the field: created 'accelerator-directory-addendum' to hold new verified entries until this is fixed; needs merging back afterward.

https://claude.ai/code/session_01WfdDLbHWF7EST34wzB6bL7

mickdarling:
## Summary

- separate structural YAML safety checks from prose/content-pattern detection
- keep size, Unicode, alias-amplification, circular-reference, YAML-bomb, schema, and Gatekeeper validation on memory saves
- stop rescanning historical memory entries with the blocking prose scanner during each append
- continue blocking malicious patterns in memory metadata, instructions, extensions, tags, sources, and other control fields\n- exempt only historical entry prose and sanitizer artifacts, not the auxiliary entry structure
- render FLAGGED entries only through sanitized or redacted content inside an untrusted-content boundary
- reject malformed or non-string sanitized representations at the display boundary and use the redacted fallback
- revalidate TRUSTED and VALIDATED entries against current scanner rules before direct output\n- sandbox stale-trust failures without mutating stored history or blocking later appends\n- add regressions using an actual blocking `exec(` pattern, save\/reload output checks, and hand-edited malformed sanitized content

## Root cause

Memory entries are stored as untrusted content and scanned asynchronously, but the persistence preflight serialized and synchronously rescanned the entire append-only memory. A scanner-rule change or false positive in any historical entry therefore made every later append fail, with no supported way to edit the old entry.

The hotfix exempts only the serialized `entries` collection from the blocking prose-pattern pass. Structural YAML protections remain mandatory for the full document, all non-entry control fields receive a second blocking content-pattern scan, and dangerous entry output remains sandboxed. FLAGGED entries use sanitized text only when it is a non-empty string distinct from the original; malformed values fail closed to a redacted placeholder.

## Validation

- focused memory API, trust-level, save-limit, and content-validator suites: 117 tests passed
- focused ESLint passed
- `tsc --noEmit` passed

Fixes #2440


mickdarling:
@codex review current head 435031c7

mickdarling:
@codex review

mickdarling:
@codex review

mickdarling:
@codex review

mickdarling:
@codex review

Issue #2652, opened 2026-09-02
Web console: show a prominent FLAGGED banner on quarantined memory entries


insomnolence:
## Summary

The web console's portfolio detail view shows a memory's stored entries verbatim, including entries whose `trustLevel` is `flagged`, without a prominent warning. The trust level is present in the data but is rendered only as an ordinary metadata item alongside the entry.

Low priority, UX only. Not a security defect: the detail route is an owner-only authoring, raw-view and download surface (`self_private`, scoped to the authenticated user), nothing forwards its response to a model, and rendering is already guarded (markdown through `sanitizeHtml`/DOMPurify, everything else through `escapeHtml`, editor content assigned via the form value). The model-context boundary in `Memory.prepareEntryForDisplay` sandboxes flagged entries independently.

## Where

- `src/web-console/ui/portfolio-detail.js`, `renderMemoryEntry` and `renderMemoryField`
- Serialized entries come from `ManagerBackedPortfolioElementStore.flatMemoryYaml`, which copies each entry intact, so `trustLevel`, `sanitizedContent` and `sanitizedPatterns` are available to the UI.

## Suggested direction

- [ ] Render a prominent FLAGGED banner on any entry whose `trustLevel` is `flagged`, stating that the content was quarantined from model context.
- [ ] Offer the `sanitizedContent` view as the default for flagged entries with an explicit "show raw" affordance, so the raw text is still available for owner repair and export.
- [ ] Surface `sanitizedPatterns` next to the banner so the owner can see why the entry was flagged.

Raised during the main→beta merge review (2026-09-02) as an inherited beta-only observation; the merge does not touch these files.

Issue #2651, opened 2026-09-02
Web console session deactivation can miss MCP-written activation records


insomnolence:
## Summary

The web console's session deactivation lookup can miss an activation record that was written by the MCP path, leaving the durable record in place while the console reports the element as not active.

Inherited beta behaviour, present at beta `17d03690`. Found during review of the main→beta merge; the file involved is not touched by that merge, so it is filed rather than fixed there.

## Where

`src/web-console/modules/activations/SessionActivationStateAdapter.ts`, `RegistrySessionActivationStateAdapter.deactivate`, matches a stored record only by

```ts
record.name === name || record.identity?.value === name
```

`upsertActivationRecord` in `src/state/activation-record-utils.ts` also matches on `filename`, so the two sides of the store do not agree on what identifies a record.

## How it is reached

1. `Container.createServerForHttpSession` attaches the child `ActivationStore` to the root `SessionActivationRegistry` state keyed by the MCP session ID, so the MCP path and the console address the same store.
2. `ElementCRUDHandler` writes an MCP activation through `AgentActivationStrategy.activationRecord`, which carries `metadata.name` (the display name) plus the storage identity.
3. The console route passes the target MCP `:session_id` through `ActivationService.deactivate`, which canonicalizes the request name with `canonicalizePortfolioElementName` before calling the adapter.
4. When the stored display name and the storage identity both differ from the canonical name, neither arm matches. The in-memory set is cleared, the durable record stays, and the console reports a successful deactivation.

Console-only flows are unaffected: `ActivationService.activate` stores `element.canonicalName`, so the name arm matches on a later console deactivate. The adapter also creates identity-absent, filename-absent records on every first activation (`recordActivation(type, displayName, storeRecord?.filename, storeRecord?.identity)` with `storeRecord` undefined), so this is not limited to legacy data.

## Suggested direction

Resolve the console's deactivation target to the same record key the activate path and `upsertActivationRecord` use, rather than comparing raw strings; or give the adapter lookup the same filename and identity arms as the shared utils.

- [ ] Cross-surface regression test: MCP-path activation with display name plus identity, console deactivate by canonical name, assert the durable record is removed.

insomnolence:
Related to #2658. The console and MCP paths resolving activation records by different rules is an instance of identity being derived per call site rather than resolved once.

Issue #2464, opened 2026-08-07
[v2.1] Complete post-release verification and branch convergence


mickdarling:
## Parent and dependency

Final phase of #2455. Begins only after #2463 publishes and deploys DollhouseMCP 2.1.

## Objective

Verify the live 2.1 release across distribution and hosted surfaces, reconcile branch metadata after release, and leave a documented one-way development flow that prevents another large two-way divergence.

## Post-release verification

- [ ] Confirm `main`, tag `v2.1.0`, npm `latest`, GitHub release, MCP Registry, MCPB/download archive, GitHub Packages, checksums/signatures, and container artifacts resolve to the intended release.
- [ ] Install from each public distribution path in clean environments and verify build info/version/channel.
- [ ] Run hosted health, readiness, OAuth, allowlist, TOTP, console, MCP transport, portfolio persistence, session, operations, and isolation smoke tests.
- [ ] Monitor deployment, Cloudflare, Caddy, container, PostgreSQL, auth, audit, and application logs for regression signals and secret leakage.
- [ ] Confirm no beta/prerelease artifact displaced a stable/latest channel incorrectly.
- [ ] Verify rollback remains functional after publication.

## Branch convergence

- [ ] Record final main, develop, beta, and hosted HTTP tips and their intended ongoing roles.
- [ ] Bring release metadata back into develop/beta through the approved reviewed process where required.
- [ ] Decide whether `codex/hosted-http-integration` is archived, retained as a historical branch, or converted to a one-way upstream feature line.
- [ ] Preserve an immutable reference to Todd's pre-reconciliation integration tip and the reconciliation merge.
- [ ] Document the future flow and back-merge policy so emergency fixes cannot silently diverge branches again.
- [ ] Close or update #2374, #2451, #2452, and the child issues only when their acceptance criteria are genuinely complete.

## Follow-up handling

- [ ] Convert every accepted non-blocking review, deployment, observability, UX, performance, or compatibility finding into a linked issue with owner/priority.
- [ ] Record deferred dependency advisories and their next review dates under #2451/#2452.
- [ ] Produce a short release-retrospective comment on #2455 covering manual work, failures, recovery, and CI/CD improvements.

## Done definition

- All public and hosted 2.1 surfaces are verified from clean state.
- Release channels are correct and traceable to one commit.
- No unexplained branch divergence remains.
- Historical authorship and rollback references are preserved.
- Follow-up work is explicitly tracked.
[...]

PR #2653, opened 2026-09-02
chore(beta): merge main at release 2.0.41 into beta

insomnolence:
## Summary

Merges `main` at release 2.0.41 into `beta` with a real merge commit, so that every 2.0.x fix
since July is inherited rather than hand-ported and the merge base between the two lines moves
from July to today. Where both lines had already fixed the same bug independently, beta's version
is kept; where beta moved code that main changed, main's change is applied at beta's new location.

Refs #2464

## Baseline

- Base branch: `beta`
- Beta tip merged onto: `17d03690d707db59dcfc178f0f2cda7b0ccde0eb`
- Main tip merged in: `5b0050bfc9f2dc42e114f61c80604f7bd1344d84` (release 2.0.41, 2026-08-26)
- Previous merge base: `0bbd0c19bc357780724fba3d46f91318872558be` (release 2.0.38, 2026-07-04)
- Merge commit: `5ef654182f33038dbdc097dfaf7caa02a36ee5ed`, parents `17d03690` and `5b0050bf`
- Main delta since the previous base: 81 commits, 93 files, releases 2.0.39, 2.0.40 and 2.0.41
- Net change on beta: 26 files, 1,623 insertions, 411 deletions

Both parents are ancestors of the merge commit. After this lands, the merge base with `main` is
`5b0050bf`, so the next `main` release merges as a small routine change.

## Scope

Of the 93 files main touched since the previous base, 43 conflicted and 50 merged automatically.
Every one of the 93 was checked by hand against both parents, because beta's refactors mean a
clean automatic merge only proves the hunks did not overlap.

- 71 files resolve to beta's existing version: beta already carried an equivalent or stronger fix,
  including the stable activation-identity chain, the collection cache path resolver, the policy
  freshness chain and the supply-chain hardening across workflows, Dockerfiles and scripts
- 9 files take main's change as-is
- 8 files take main's change relocated into beta's structure, principally the agent read chain
  into the element loader and agent state store, and the execution-policy recovery chain into the
  agent execution handler
- 5 files are hybrids of both

Two files main added are not carried: a cache-path helper and its test, because beta's
platform-aware data-directory resolver already provides the same behaviour with stricter override
handling and has its own tests.

Files this merge touched are left with zero SonarLint issues; that cleanup is part of the merge
commit.

## Behaviour changes on beta

**Agents.** Ordinary agent reads stay tolerant of unrelated lookup failures while recovery reads
are strict: definition-only candidate reads do not hydrate or cache, strict recovery re-reads
durable state for the requested identity, and missing storage is distinguished from unavailable
storage. Execution identities are sanitized before normalization, so `foo&bar` and `foo-bar` no
longer collide on one execution key. Stale-policy recovery holds its execution-generation
observation across the revalidating read.

**Gatekeeper.** A denied operation's error now carries the actionable suggestion the policy
resolver produced.

**Portfolio.** Strict listing callers receive filesystem access failures instead of an empty
result.

**Collection cache.** Two redundant success log lines are removed.

Everything else main changed in this window either already shipped on beta in equivalent or
stronger form or is documentation.

## Validation

At the merge commit `5ef65418`, dispatched on this branch before the PR was opened:

- Core Build & Test (run 33681267892): unit green on Ubuntu, macOS and Windows; PostgreSQL
  Integration 126 suites, 2,653 tests. The advisory performance step recorded one wall-clock
  timing miss on macOS; it does not gate.
- Extended Node Compatibility (run 33681270299): green on all six OS and Node 20/22 jobs.
- CodeQL CLI on an exact archive of the commit: 86 queries, findings identical to beta's baseline
  by fingerprint, none in the touched files.

Locally on the same tree: lint zero warnings; build; unit 613 suites, 13,788 tests; integration
124 suites, 2,512 tests; fixture E2E 186/186; console E2E 139/139; compiled browser E2E 55/55;
SonarLint 0 issues across the 28 touched files.

Two inherited observations surfaced during review and are filed rather than changed here, since
neither is in a file this merge touches: #2651 and #2652.

## Reviewing

To see what the merge changed relative to either parent:

```
git diff 17d03690..5ef65418 -- <path>     # change on beta
git diff 5b0050bf..5ef65418 -- <path>     # change relative to main
```

A file absent from the first diff is unchanged for beta users regardless of what main did to it.

## Scope controls

- No version, release-channel or publication change; `package.json` version is unchanged.
- No dependency or lockfile change; `package.json`, `package-lock.json` and
  `packages/safety/package-lock.json` are byte-identical to `17d03690`.
- No workflow change; no file under `.github/` differs from `17d03690`.
- No test weakened. Main's test intent is preserved where a test was adapted to beta's structure;
  three interleaving tests that beta's per-agent serialization makes unreachable were replaced by
  tests of the serialization contract.
- Merge commit only. Do not squash or rebase.


mickdarling (commented):
### Additional local identity-integrity review at `d07690d0`

I ran a focused, read-only review against the exact current head, including temporary regression probes that were removed after execution. The existing focused unit suites pass (142 tests), and the existing PostgreSQL RLS/agent-state suites pass (23 tests), but those suites do not cover two important matrix cells. Both new probes failed reproducibly.

#### P1: Database storage indexes are shared across dynamic users

`AbstractDatabaseStorageLayer` holds `nameToIdMap`, `idToNameMap`, `lastScanTimestamp`, and `scanCompleted` as instance-wide state, while `getCurrentUserId()` intentionally changes per request. The root element managers and their storage layers are singleton-scoped, so one user's index remains active when the same manager serves another user.

Reproduction:

1. A shared layer running as user A writes a public row named `shared-exact-name`; its map records A's UUID.
2. A separate layer writes user B's private row with the same exact name.
3. Switch the shared layer's resolver to user B and call `getPathByName('shared-exact-name')`.
4. Expected B's UUID; received A's UUID.

RLS still protects private rows, but it cannot fix the identity error. With a public A row, user B can resolve and load A's public definition instead of B's own exact-name definition. For agents, that can select the wrong instructions or policy. The index and scan state need to be namespaced by current user (and element type), or each user must receive an isolated storage-layer instance.

#### P1: Canonical execution keys merge distinct legal database agents

`AgentExecutionHandler.executionKey()` and `AgentManager`'s state/generation locks use `canonicalizeExecutionName()`. PostgreSQL permits distinct exact names such as `Case_Sensitive-Agent` and `case-sensitive-agent`, but both canonicalize to the same key.

A focused handler probe executed those two agents in one session. Expected two execution entries; received one. The second execution merged both goal IDs into the shared entry and replaced its `name`, Gatekeeper metadata, and resilience policy with the second agent's values. Subsequent step, completion, abort, notification, and policy operations can therefore act through the sibling's lifecycle entry.

Execution ownership should be keyed by the resolved durable row UUID (plus user/session scope), not by a lossy display-name canonicalization. Canonicalization can remain a lookup aid, but it cannot be the identity authority when canonically colliding rows are legal.

These are separate from Codex's three current-head P1 findings concerning stale delete aliases, fresh UUID deletion, and cache recovery collisions. No repository or PR code was changed by this review.

Issue #2677, opened 2026-09-11
[Beta EPIC] Invite-only onboarding with email claim and required GitHub login


mickdarling:
## Decision

The private hosted beta will use:

- **Email as the invitation, inbox-verification, notification, and controlled-recovery channel.**
- **GitHub OAuth as the required normal login pipeline.**
- **No Dollhouse password in the beta invitation flow.**
- **No automatic account linking by matching email addresses.**
- **No coupling between GitHub authentication and the optional GitHub portfolio integration.**

Every beta tester must have a GitHub account. The invitation email does not need to match the email recorded on GitHub because the account binding is proven by possession of the bespoke invitation plus GitHub authentication in the same restricted onboarding session.

## Delivery target

- All implementation work for this issue targets the `beta` branch.
- Pull requests must use `beta` as their base branch.
- Acceptance requires validation from the `beta` branch deployment; a change merged only to `main` or another integration branch does not complete this issue.

## Invitation expiration policy

- The bespoke invitation/claim link defaults to **24 hours**.
- Operators may configure the invitation TTL below 24 hours, subject to a separate documented safe lower bound.
- The existing 15-minute generic magic-link TTL must not be inherited or reused for beta invitations.
- Keep invitation TTL separate from the restricted onboarding-session TTL, GitHub OAuth state TTL, verification-code TTL, and recovery-token TTL; those security credentials may remain short-lived.
- The administrator sees the effective lifetime before sending, and the email plus console show the exact expiration timestamp and human-readable remaining time.
- The upper bound must be explicit, validated, and documented rather than unbounded.

## Context

The live hosted beta cannot currently onboard outside testers smoothly:

- Public console sign-in goes directly to GitHub OAuth.
- Creating a console user can produce a local-account invite URL but does not deliver the intended GitHub-required enrollment experience.
- The existing local-account claim path asks for a password and persists `emailVerified: false`.
- Todd's beta work includes a Nodemailer/SMTP sender and a magic-link method, but hosted deployment automation does not provision the SMTP values and invitation issuance is not wired to that sender.
- The v2.0.42 commercial/enterprise verification Worker already sends through the existing Resend account.
- The codebase supports multiple authentication identities per user and administrator-managed linking, but it does not expose a secure self-service “attach this GitHub login identity to my invited account” flow.
- Ordinary GitHub sign-in can provision a separate principal for an unlinked provider subject, so simply sending an email and then asking the tester to sign in with GitHub can create the wrong account.

[...]

Issue #2670, opened 2026-09-11
[Beta console] Explain administrative roles in invite and user-detail flows


mickdarling:
## Context

The beta console asks administrators to assign roles without explaining what those roles mean.

The invite dialog renders only role labels. The user-detail drawer adds raw capability names, but those strings do not explain the practical authority, sensitive actions, intended audience, or consequences of each role. This makes it easy to over-grant access.

Current roles include `admin`, `account_admin`, `operator`, `auditor`, and `security_admin`.

## Delivery target

- All implementation work for this issue targets the `beta` branch.
- Pull requests must use `beta` as their base branch.
- Acceptance requires validation from the `beta` branch deployment; a change merged only to `main` or another integration branch does not complete this issue.

## Desired behavior

Provide a single role-description catalog that the invite flow and user-detail role editor both render. Each role should have a plain-language summary, important capabilities, sensitive powers, prerequisites such as TOTP elevation, and any role-management limits.

## Acceptance criteria

- [ ] Every assignable role has a human-readable name and concise description.
- [ ] The invite dialog shows the description before a role is selected.
- [ ] The user-detail Roles section shows the same description and the effective capabilities.
- [ ] The UI explains what an account with no administrative roles can do.
- [ ] Locked/unassignable roles explain which authority is required instead of appearing as unexplained disabled controls.
- [ ] Role descriptions come from one authoritative definition so UI surfaces cannot drift.
- [ ] Descriptions are keyboard-accessible and do not rely only on hover tooltips.
- [ ] Tests cover catalog completeness and rendering in both account flows.

## Branch acceptance

This beta dependency is not complete until its UI and API changes are merged to and validated from the `beta` branch.

## Related

- #2247
- #2266

mickdarling:
Implementation merged into beta in #2693 (ae424d78344be2d0e90ad2b8e84a0c7bae331e6a), with all 23 checks and final-head Codex/Claude reviews clear. The authoritative role catalog, both form renderings, disabled-role explanations, and keyboard accessibility are implemented and tested. Keeping this issue open for the required hosted-beta deployment validation; merging alone does not claim that acceptance step.

PR #2693, opened 2026-09-20
feat(console): explain administrative roles in account flows

mickdarling:
Administrators currently choose roles without a clear explanation of their authority. The invite dialog and user role editor now use one server-owned description catalog to show each role's purpose, sensitive powers, effective capabilities, TOTP requirements, and assignment limits. Unavailable roles remain visible with a keyboard-readable explanation, and the invite dialog scrolls on smaller screens.

The change adds descriptive metadata to the existing role catalog and keeps authorization enforced by the existing backend. It does not enable the new invitation lifecycle.

Validation: build, full lint, and 10 focused catalog/rendering tests passed, including both invitation and role-edit modes, disabled-role explanations, selected state, and HTML escaping.

Refs #2670 and #2677. Keep #2670 open until validation on the deployed beta, as required by its acceptance criteria.

Merge gate: all CI and Codex/Claude reviews must cover the final head; review findings must be resolved or explicitly tracked as nonblocking follow-ups. The user has authorized a merge into beta once these gates pass, including the established admin override for the unavailable second human reviewer.


mickdarling:
@codex review

mickdarling:
@claude Please recheck the two proposed pre-merge findings against the same exact head `d83663dab8ccdfd2decada73120ebff46a90c527` and state whether any blocking findings remain. Verify HEAD before reviewing.

- The authority reason is assembled as plain text, then escaped at its only HTML sink: `<span class="ua-role-opt-reason">${escapeHtml(reason)}</span>`. The existing hostile-capabilities test runs without actor capabilities, so it exercises this locked-role reason path as well as the effective-capabilities path and asserts there are no script/image elements. There is no unescaped capability interpolation into HTML.
- Labels are not ordinarily Tab stops, and a disabled checkbox is removed from sequential keyboard focus. The explicit `tabindex="0"` is only on locked-role labels, so keyboard users can reach and read their visible explanations. Enabled labels get no tabindex. Removing it would make that explanation unreachable by Tab; adding aria-disabled alone would not provide focusability.
- Removing opacity is intentional to preserve readable contrast for the explanation. The native disabled checkbox, explanatory text, and cursor distinguish locked controls.

The account-administration prerequisite matches the current backend policy; changing that policy is outside this copy/UI PR. Missing optional descriptive metadata has a visible fallback, and HTML text is escaped. There is no barrel export requirement here.

Issue #2681, opened 2026-09-11
[Beta auth] Activate invited accounts atomically after GitHub identity enrollment


mickdarling:
## Parent

- #2677 — invite-only onboarding and GitHub-required account claiming

## Delivery target

- All implementation work for this issue targets the `beta` branch.
- Pull requests must use `beta` as their base branch.
- Acceptance requires validation from the `beta` branch deployment; a change merged only to `main` or another integration branch does not complete this issue.

## Depends on

- #2678 — durable invitation and pending-account lifecycle
- #2680 — restricted invitation-claim onboarding session
- #2679 — GitHub login-identity linking
- #2670 for authoritative role descriptions and assignment semantics

## Problem

Verifying an email, attaching a GitHub identity, granting roles, consuming an invitation, and issuing an active session span several pieces of state. If those mutations are performed independently, failures or retries can produce partially active accounts, usable invitations after activation, orphan identities, duplicate principals, or users with the wrong roles.

The system also needs an enforceable rule that private-beta users authenticate through GitHub after onboarding. Merely hiding alternative login buttons in the frontend does not enforce that rule.

## Goal

Create one transactional activation boundary that converts a restricted, email-verified pending account into an active GitHub-authenticated account only when every required invariant is satisfied.

## Expiration boundary

The activation transaction uses the invitation's persisted expiration. Hosted-beta invitations default to 24 hours and may be deliberately configured shorter within documented safe bounds, but the account must still fail activation if that invitation expires while GitHub OAuth is in progress. OAuth state and restricted sessions remain independently short-lived and do not extend invitation eligibility.

## Activation preconditions

Activation must verify inside the transaction:

- Account state is `pending_activation`.
- Invitation is the current `pending` generation.
- Invitation has not expired or been revoked/superseded.
- Invitation normalized email matches the verified-email assertion.
- The current onboarding session is bound to the invitation and account.
- A unique GitHub login identity is attached to this account.
- The GitHub identity is not attached to another account.
- Intended roles remain valid and assignable under current policy.
- The inviting/approving administrator's decision has not been revoked where policy requires it.
- Private-beta capacity or policy gates still permit activation, if configured.

## Atomic mutation

In a single database transaction:

1. Lock or conditionally update the pending account and current invitation generation.
2. Confirm the linked GitHub identity.
3. Apply the approved initial roles/access.
4. Mark account `active`.
5. Mark invitation `accepted` with timestamp.
6. Make every invitation credential unusable.
[...]

PR #2701, opened 2026-09-20
feat(auth): add fail-closed identity transaction try-locks

mickdarling:
Authentication mutations that already hold account locks need a fail-closed way to acquire the existing principal and allowlist advisory locks without waiting behind another identity writer. Add one reusable transaction helper that uses the existing namespaces and sorted keys, and raises PostgreSQL `40001` on contention so partial writes and locks cannot commit even if JavaScript catches the error.

Existing blocking helpers and sign-in/deletion policies remain unchanged. This is a bounded prerequisite for #2681, not activation: the included lock audit explicitly identifies later identity-row, role and global audit-chain locks that the activation transaction must also address before it can be enabled. No OAuth, session, route or activation behavior is introduced.

Validation: build, full lint, script typecheck, 146 existing unit tests, and four real PostgreSQL 17 tests against an isolated disposable database. Integration coverage includes actual sign-in provisioning and deletion interleavings, partial-lock/write rollback, caught transaction errors, and deletion tombstone precedence on fresh sign-in.


mickdarling:
@codex review

mickdarling:
@codex review

Fixed optional PostgreSQL availability in 68137d5a. The spec now follows isDatabaseAvailable(), closes its helper pool, and skips its bodies when PostgreSQL is optional/unavailable; DOLLHOUSE_REQUIRE_TEST_DATABASE=1 still fails closed. Verified an actual unreachable app URL at 127.0.0.1:1 with isolated no-op global setup/teardown, then reran all four tests against required disposable PostgreSQL. Focused lint passed.

mickdarling:
Final head `68137d5a24438e3412474230a8a45ee477b88281` has all 23 checks passing and clean exact-head Codex/Claude reviews. The optional-PostgreSQL review finding is fixed and resolved. The unrelated Windows TLS test passed one targeted retry; its reliability follow-up is #2708.

Validated the exact combination with current beta `65c6bd91` in an isolated checkout: clean merge, full build, 32 database unit tests, and all four real PostgreSQL contention/rollback tests passed. This is an independent helper; it does not enable activation or change existing blocking callers.

Using the owner's admin-merge authorization for the unavailable human approval and strict base freshness on this tested combination. No failing or pending check is overridden.

Issue #2800, opened 2026-09-22
Ensemble activation fails for personas not in the element cache


insomnolence:
**Problem**
Activating an ensemble skips its persona member with `Persona '<name>' not found` whenever the persona isn't cached. Other members activate normally, and the persona can be listed, viewed and activated directly. On database-backed servers this happens every time (#2799). In file mode it happens after a restart, after cache expiry, or after eviction.

**Reproduce:** on a database-backed server, activate any ensemble with a persona member. The result shows `Failed: 1`.

**Cause**
`Ensemble.findElementInManager()` (`src/elements/ensembles/Ensemble.ts`) resolves personas with the cache-only `PersonaManager.findPersona()`. Every other persona lookup uses `findPersonaAsync()`, which falls back to storage. Unit tests mock `findPersona`, and the only integration test creates the persona right before activating, so it's always cached.

**Test first** (fails on `beta`; add to `tests/integration/mcp-aql/ensemble-activation-registration.test.ts`)
Create a persona and an ensemble with it as `primary`. Call `personaManager.clearCache()`, activate the ensemble, and assert no failed elements and that the persona is active.

**Possible fix**
Use `await manager.findPersonaAsync(name)`, or drop the persona special case and use `findByName()` like the other types. A few lines; update the mocks in `tests/unit/elements/ensembles/Ensemble.test.ts`.

**Related:** #2799

insomnolence:
Fixed by #2802, merged into `beta` in f817c01d.

Persona lookups that used the cache-only `findPersona()` now use `findPersonaAsync()`, which checks the cache and falls back to storage:

- Ensemble activation (`Ensemble.findElementInManager()`)
- Ensemble create/edit type resolution (`resolveElementTypes()`), including persona/skill ambiguity
- Persona deactivation (`PersonaActivationStrategy.deactivate()`)
- Ensemble deactivation of persona members; a member that can't be resolved at all is logged and skipped instead of aborting
- Activation-record filenames and `PersonaManager.validatePersona()`

Regression tests in `tests/integration/mcp-aql/persona-cold-cache.test.ts` and `tests/integration/mcp-aql/ensemble-activation-registration.test.ts` clear the persona cache before exercising each path; they failed before the fix and pass after.

Closing manually because merges into `beta` don't auto-close issues. The underlying database cache wipe is tracked in #2799.

Issue #2801, opened 2026-09-22
Database mode: findByName() never reads the element cache


insomnolence:
**Problem**
In database mode, every name lookup (details, activation, `findPersonaAsync`, policy member resolution) reads Postgres even when the element is cached. Loads write to the cache, but name lookups never read it back.

**Reproduce:** the server log shows `[PersonaManager] Cache miss recovered from disk` on repeated lookups of the same persona.

**Cause**
In `ElementResolver.findByName()` (`src/elements/base/ElementResolver.ts`):
1. `findInCache()` builds file paths from the name. Database-mode elements are cached under their row UUID, so this never hits.
2. The name is resolved to a UUID via `getPathByName()`, then `host.load(uuid)` is called without checking the cache.
3. `ElementLoader.load()` always reads storage.

**Tests first** (fail on `beta`; new unit test with a stubbed writable storage layer, or database integration)
1. Load `X`, call `findByName('X')` twice, and assert storage `readContent` runs once.
2. Update `X`'s row, refresh, call `findByName('X')`, and assert the new content is returned.

**Possible fix**
After `getPathByName()`, return `cache.getCachedByPath(uuid)` if present, and call `load()` only on a miss. A few lines. **Depends on #2799**: cached reads are only safe once refresh evicts precisely. Test 2 guards that.

**Related:** #2799, #2800

insomnolence:
Blocked on the cache-coherence redesign in #2799; this fix is only safe once cached entries are version-validated.

Issue #2799, opened 2026-09-22
Element cache coherence: DB refresh wipes the cache every op; selective eviction needs a redesign (both storage modes)


insomnolence:
**Problem**
In database mode, `refreshIndex()` / `scanAndEvict()` evicts every cached element of that type, even when nothing changed. Gatekeeper enforcement refreshes personas, skills and ensembles before each MCP-AQL operation, so on database-backed servers those caches are almost always empty. File mode isn't affected; its diff is based on modification times.

**Reproduce:** on a database-backed server, run any two MCP-AQL operations, then `get_cache_budget_report`. Personas, skills and ensembles show 0 entries, while agents and templates stay cached.

**Cause**
`ElementListOperations.scanAndEvict()` (`src/elements/base/ElementListOperations.ts`) calls `invalidate()` → `scan()` → uncaches `modified` + `removed`. `AbstractDatabaseStorageLayer.invalidate()` (`src/storage/AbstractDatabaseStorageLayer.ts`) nulls `lastScanTimestamp`, so the next scan is full, and `processSubsequentScanRows()` marks every known row `modified`. The index keeps no per-row version.

**Tests first** (fail on `beta`; add to `tests/integration/database/DatabaseStorageLayer.test.ts`, needs Postgres)
1. Load `X`, refresh without changes, and assert `X` is still cached and `modified` is empty.
2. Load `X` and `Y`, update `X`, refresh, and assert only `X` is evicted.
3. Load `X`, delete its row, refresh, and assert `X` is in `removed`.
4. In `tests/integration/database/mcp-database-e2e.test.ts`, run two back-to-back operations and assert the persona cache survives.

**Possible fix**
Store `updatedAt` per id in `DatabaseIndexState`, and report `modified` only when it differs. Keep `invalidate()` forcing a full query so deletions are still detected. About 20–30 lines in one file.

**Related:** #2618, #2634

insomnolence:
**Status: needs a cache-coherence redesign. A diff-only fix would be a regression.**

We implemented the fix proposed above (track `updated_at` per row and report `modified` only on a version change), plus follow-ups from review:
- microsecond-precision versions;
- a guard so an in-flight load can't cache content an eviction already invalidated;
- eviction of rows first seen as `added`;
- handling for renames and same-name replacement.

It restores caching in database mode and passes the tests listed in this issue.

[...]

PR #2802, opened 2026-09-23
fix(personas): resolve uncached personas via storage-backed lookup

insomnolence:
## Summary

Fixes #2800.

Several persona lookups used `PersonaManager.findPersona()`, which is synchronous and only searches the in-memory element cache. When a persona existed in storage but wasn't cached, those paths treated it as missing. On database-backed servers the persona cache is usually cold (see #2799), so ensembles consistently activated without their persona member, and persona deactivation failed with "not found".

Each of these paths now uses `findPersonaAsync()`, which checks the cache first and falls back to storage:

- **Ensemble activation:** `Ensemble.findElementInManager()` now loads persona members that aren't cached.
- **Ensemble create/edit type resolution:** `resolveElementTypes()` resolves untyped persona members correctly, and reports persona/skill name collisions as ambiguous instead of silently picking the skill.
- **Persona deactivation:** `PersonaActivationStrategy.deactivate()` looks up uncached personas instead of rejecting them. It already returned a Promise, so callers are unchanged.
- **Ensemble deactivation:** persona members are loaded from storage before deactivation, so they no longer stay active on a cold cache; a member that can't be resolved at all is logged and skipped instead of aborting the deactivation.
- **Activation records:** the persisted activation filename no longer comes back empty for uncached personas.
- **Persona validation (legacy path):** `PersonaManager.validatePersona()` is now async, and its only caller awaits it.

`PersonaManager.create()` keeps its synchronous duplicate check because it runs immediately after a full `list()`, so the cache is populated. The synchronous active-persona accessors are unchanged.

## Tests

New regression tests use the real `PersonaManager` and storage, clear the persona cache, then exercise each path. All failed before the fix and pass after:

- `tests/integration/mcp-aql/persona-cold-cache.test.ts` (new): type resolution, persona/skill ambiguity, validation, deactivation, and the activation record.
- `tests/integration/mcp-aql/ensemble-activation-registration.test.ts`: ensemble activation with an uncached persona.

Lint and build pass; the full unit and integration suites pass.

## Related

- #2799: database refresh wipes the element cache, which makes this failure constant in database mode.
- #2801: database-mode `findByName()` doesn't read the element cache.

Note: this targets `beta`, so `Fixes #2800` won't close the issue automatically on merge.


insomnolence:
Re the automated review: `PersonaHandler.validatePersona()`'s pre-check calls `personaManager.find(predicate)`, which is async and runs a full `list()` from storage (`BaseElementManager.find` → `ElementResolver.find`), so it doesn't depend on the cache. `PersonaManager.reload()` doesn't call `deactivatePersona()`; the two by-name callers (`PersonaActivationStrategy.deactivate` and ensemble member deactivation) now both resolve the persona via `findPersonaAsync()` first. The silent `catch` in `findMatchingTypes` matches existing behavior and is out of scope for this fix.

mickdarling (approved):
Reviewed commit a75ce78a878f30a1e027a5783025d0e7cfd58d39. No blocking findings.

Storage-backed persona resolution is awaited on the affected validation, activation, and ensemble deactivation paths. The direct handler deletion retains PersonaManager's deactivation, reload, audit, and sanitized-error behavior.

Claude review disposition: the cold-cache validation concern is a false positive: PersonaHandler awaits inherited find(predicate), whose resolver awaits list() before scanning, then awaits manager validation. The reload/deactivate call-path concern does not match the implementation; the affected by-name deactivation callers load the persona first. The silent resolver catch is preexisting, not introduced here.

Validation: all 24 checks pass on this exact head, including PostgreSQL integration, security, and SonarCloud. Independent focused local validation passed 117 unit tests and 11 integration tests. Local PostgreSQL was unavailable; PostgreSQL coverage is from CI.

The broader cache follow-ups tracked in #2799 and #2801 remain separate. Approved for merge into beta.

